PR #3, opened 2025-08-07
Merge Script Tools API and WebMC explainers [Part I]

bwalderman:
This is a first pass at unifying Microsoft's [Web Model Context](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/WebModelContext/explainer.md) and Google [Script Tools](https://github.com/explainers-by-googlers/script-tools) explainers.

## Summary of changes

- Merged in explainer content from Script Tools API
  - Background and motivation
  - Goals and non-goals
  - Example use cases
  - Prior art
  - Images
- Updated new Prior Art section with acknowledgement of pre-existing WebMCP from the MCP-B project
- Split out API proposal into a separate document to keep size down
  - Currently contains Web Model Context API design and Stamp database walkthrough.

## Follow-ups

The proposal.md still contains just Web Model Context API stuff, more or less verbatim. Still need to converge on a proposed API design that takes Script Tools, WebMCP (MCP-B), and other prior art into account. Keeping that out of this PR to avoid it becoming too large and since the API is still being discussed.

+khushalsagar


docs/proposal.md
```diff
@@ -0,0 +1,258 @@
...
+
+One proposal was to expose the WebMCP API only in service workers and let the service worker post messages to individual client windows/tabs as needed in order to update UI. This would have complicated the architecture and required web developers to add a service worker. This would also have required the Session concept described earlier to help the service worker differentiate between agents that are connected to different windows and dispatch requests from a particular agent to the correct window.
+
+For long-running, batched, or expensive tool calls, we expect web developers will dynamically update their UI when these are taking place to temporarily cede control to the agent (e.g. disable or remove human form inputs, indicate via UI that an agent is in control), and take advantage of dedicated workers as needed to offload expensive operations. This can be achieved with existing dedicated or shared workers.
```
anssiko:
```suggestion
For long-running, batched, or expensive tool calls, we expect web developers will dynamically update their UI when these are taking place to temporarily cede control to the agent (e.g. disable or remove human form inputs, indicate via UI that an agent is in control), and take advantage of dedicated workers as needed to offload expensive operations. This can be achieved with existing dedicated or shared workers.

## Acknowledgments

Many thanks to [Alex Nahas](https://github.com/MiguelsPizza) for sharing related [implementation experience](https://github.com/MiguelsPizza/WebMCP). 
```

I'd like to include these acknowledgments to @miguelspizza and his project, either in the context of this PR that implements the rename, or in a separate PR.

anssiko (approved):
Signing off for the general direction with a non-blocking suggestion to include acknowledgments.

docs/explainer.md
```diff
@@ -1,345 +1,609 @@
...
 
-Primary goals of the Web Model Context API:
+WebMCP is a proposal for a web API that enables web pages to provide agent-specific paths in their UI. With WebMCP, agent-service interaction takes place _via app-controlled UI_, providing a shared context available to app, agent, and user. In contrast to backend integrations, WebMCP tools are available to an agent only once it has loaded a page and they execute on the client. Page content and actuation remain available to the agent (and the user) but the agent also has access to tools which it can use to achieve its goal more directly.
 
-- **Enable human-in-the-loop workflows**: Support scenarios where users work directly through delegating tasks to AI agents or assistive technologies while maintaining much of the visibility and control over the web page(s).
-- **Simplify AI agent integration**: Allow AI agents to interact with web sites through well-defined tools defined in JavaScript rather than screen parsing, automation, or writing back-end code.
+![A diagram showing an agent communicating with a third-party service via script tools running in a live web page](../content/explainer_st.svg)
```
khushalsagar: Let's use WebMCP now for consistency.

docs/explainer.md
```diff
@@ -1,345 +1,609 @@
...
+
+In contrast, in a backend integration, the agent-service interaction takes place directly, without an associated UI. If
+a UI is required it must be provided by the agent itself or somehow connected to an existing UI manually:
+
+The expected flow using browser agents and Script Tools:
```
khushalsagar: same here

docs/explainer.md
```diff
@@ -1,345 +1,609 @@
...
 - **Headless browsing scenarios**: While it may be possible to use this API for headless or server-to-server interactions where no human is present to observe progress, this is not a current goal. Headless scenarios create many questions like the launching of browsers and profile considerations.
 - **Autonomous agent workflows**: The API is not intended for fully autonomous agents operating without human oversight, or where a browser UI is not required. This task is likely better suited to existing protocols like [A2A](https://a2aproject.github.io/A2A/latest/).
-- **Replacement of existing protocols**: Web Model Context works with existing protocols like MCP and is not a replacement of existing protocols.
+- **Replacement of backend integrations**: WebMCP works with existing protocols like MCP and is not a replacement of existing protocols.
+- **Replace human interfaces**: The human web interface remains primary; agent tools augment rather than replace user interaction.
+- **Enable / influence discoverability of sites to agents**
+
+## Use Cases
+
+The use cases for script tools are ones in which the user is collaborating with the agent, rather than completely
```
khushalsagar: same here

docs/explainer.md
```diff
@@ -1,345 +1,609 @@
...
-                }
-            }
-        ]
-    });
-}
-```
+## Assumptions
 
-Now the author needs to implement the tool. The tool needs to update the stamp database, and refresh the UI to reflect the change to the database. Since the code to do this is already available in the `addStamp()` function written earlier, the tool implementation is very simple and just needs to call this helper when an "add-stamp" tool call is received. After calling the helper, the tool needs to signal completion and should also provide some sort of feedback to the client application that requested the tool call. It returns a text message indicating the stamp was added:
+* For many sites wanting to integrate with agents quickly - augmenting their existing UI with script tools will be
```
khushalsagar: WebMCP here as well.

docs/explainer.md
```diff
@@ -1,345 +1,609 @@
...
+```
 
-### Permissions
+Unlike OpenAPI, MCP is transport-agnostic. It comes with two built in transports: stdio which uses the systems standard
+input/output, well suited for local communication between apps, and Server-Sent Events (SSE) which uses HTTP commands
+for remote execution.
+
+### WebMCP (MCP-B)
+
+[MCP-B](https://mcp-b.ai/), or Model Context Protocol for the Browser, is an open source project found on GitHub [here](https://github.com/MiguelsPizza/WebMCP) and has much the same motivation and solution as described in this proposal. MCP-B's underlying protocol, also named WebMCP, extends MCP with tab transports that allow in-page communicate between a website's MCP server and any client in the same tab. It also extends MCP with extension transports that use Chromium's runtime messaging to make a website's MCP server available to other extension components within the browser (background, sidebar, popup), and to other external MCP clients running on the same machine. MCP-B enables tools from different sites to work together, and for sites to cache tools so that they are discoverable even if the browser isn't currently navigated to the site.
```
khushalsagar:
```suggestion
[MCP-B](https://mcp-b.ai/), or Model Context Protocol for the Browser, is an open source project found on GitHub [here](https://github.com/MiguelsPizza/WebMCP) and has much the same motivation and solution as described in this proposal. MCP-B's underlying protocol, also named WebMCP, extends MCP with tab transports that allow in-page communication between a website's MCP server and any client in the same tab. It also extends MCP with extension transports that use Chromium's runtime messaging to make a website's MCP server available to other extension components within the browser (background, sidebar, popup), and to other external MCP clients running on the same machine. MCP-B enables tools from different sites to work together, and for sites to cache tools so that they are discoverable even if the browser isn't currently navigated to the site.
```

docs/proposal.md
```diff
@@ -0,0 +1,262 @@
...
+## API
+
+The `window.agent` interface is introduced to represent an abstract AI agent that is connected to the page and uses the page's context. The `agent` object has a single method `provideContext` that's used to update the context (currently just tools) available to the agent. The method takes an object with a `tools` property which is a list of tool descriptors. The tool descriptors look as shown in this example below, which aligns with the Prompt API's [tool use](https://github.com/webmachinelearning/prompt-api#tool-use) specification, and other libraries like the MCP SDK:
+
+```js
+// Declare tool schema and implementation functions.
+window.agent.provideContext({
+    tools: [
+        {
+            name: "add-todo",
```
khushalsagar, line 53: Is the name expected to be unique across the list of tools? If yes, can you document that.

docs/proposal.md
```diff
@@ -0,0 +1,262 @@
...
+- Allows additional context different discovery mechanisms without rendering a page.
+
+**Disadvantages:**
+
+- Slightly harder to keep definition and implementation in sync.
+- Potentially large switch-case in event handler.
+
+### Recommendation
+
+A **hybrid** approach of both of the examples above is recommended as this would make it easy for web developers to get started adding tools to their page, while leaving open the possibility of manifest-based approaches in the future. To implement this hybrid approach, a `"toolcall"` event is dispatched on every incoming tool call _before_ executing the tool's `execute` function. The event handler can handle the tool call by calling the event's `preventDefault()` method, and then responding to the agent with `respondWith()` as shown above. If the event handle does not call `preventDefault()` then the browser's default behavior for tool calls will occur. The `execute` function for the requested tool is called. If a tool with the requested name does not exist, then the browser responds to the agent with an error.
```
khushalsagar: We need a deeper discussion for the API but curious as I'm reading this. I'm not following if the recommendation is for the json manifest based declaration or the `provideContext` API based (which includes the execute function).

docs/proposal.md
```diff
@@ -0,0 +1,262 @@
...
+## Understanding WebMCP
+
+Only a top-level browsing context, such as a browser tab can be a model context provider. A page calls the WebMCP API's methods to register tools with the browser. An agent requires some information from the tool in order to use it. A simple, common subset emerges from [existing AI integration APIs](explainer.md#prior-art):
+
+* A natural language description of the tool / function
+* For each parameter:
+  * A natural language description of the parameter
+  * The expected type (e.g. Number, String, Enum, etc)
+  * Any restrictions on the parameter (e.g. integers greater than 0)
+* A JS callback function that implementings the tool and returns a result
```
bokand:
```suggestion
* A JS callback function that implements the tool and returns a result
```

docs/proposal.md
```diff
@@ -0,0 +1,262 @@
...
+
+When an agent that is connected to the page sends a tool call, the JavaScript callback is invoked, where the page can handle the tool call and respond to the agent. Simple applications can handle tool calls entirely in page script, but more complex applications may choose to delegate computationally heavy operations to workers and respond to the agent asynchronously.
```
bokand: Might be good to make explicit that this can happen asynchronously. i.e. the tool call isn't complete until the `execute` promise resolves.

docs/proposal.md
```diff
@@ -0,0 +1,262 @@
...
+
+Handling tool cools in the main thread with the option of delegating to workers serves a few purposes:
```
bokand: Have you considered delegating to a real MCP server as well? This is one option we've considered as concern about "duplicated effort between this and MCP" was a commonly heard concern.

docs/proposal.md
```diff
@@ -0,0 +1,262 @@
...
+
+One proposal was to expose the WebMCP API only in service workers and let the service worker post messages to individual client windows/tabs as needed in order to update UI. This would have complicated the architecture and required web developers to add a service worker. This would also have required the Session concept described earlier to help the service worker differentiate between agents that are connected to different windows and dispatch requests from a particular agent to the correct window.
+
+For long-running, batched, or expensive tool calls, we expect web developers will dynamically update their UI when these are taking place to temporarily cede control to the agent (e.g. disable or remove human form inputs, indicate via UI that an agent is in control), and take advantage of dedicated workers as needed to offload expensive operations. This can be achieved with existing dedicated or shared workers.
+
+## Acknowledgments
+
+Many thanks to [Alex Nahas](https://github.com/MiguelsPizza) for sharing related [implementation experience](https://github.com/MiguelsPizza/WebMCP).
```
anssiko:
```suggestion
Many thanks to [Alex Nahas](https://github.com/MiguelsPizza) and [Jason McGhee](https://github.com/jasonjmcghee/) for sharing related [implementation](https://github.com/MiguelsPizza/WebMCP) [experience](https://github.com/jasonjmcghee/WebMCP).
```

I invited @jasonjmcghee to this group (welcome!) and with his permission I'd like to acknowledge also [his WebMCP project](https://github.com/jasonjmcghee/WebMCP) that prototyped the concept using a websocket server as a bridge. I'm sure Jason is happy to answer questions related to his implementation as they emerge.

docs/proposal.md
```diff
@@ -0,0 +1,262 @@
...
+});
+```
+
+Tool calls are handled as events. Since event handler functions can't respond to the agent by returning a value directly, the `'toolcall'` event object has a `respondWith()` method that needs to be called to signal completion and respond to the agent. This is based on the existing service worker `'fetch'` event.
+
+**Advantages:**
+
+- Allows additional context different discovery mechanisms without rendering a page.
+
+**Disadvantages:**
```
bokand, line 128: Another disadvantage compared to the imperative form is that these cannot be context-dependent - i.e. the imperative form allows you to reset the set of available tools by calling `provideContext`. The declarative form is effectively "always available".

docs/proposal.md
bokand:
```suggestion
A **hybrid** approach of both of the examples above is recommended as this would make it easy for web developers to get started adding tools to their page, while leaving open the possibility of manifest-based approaches in the future. To implement this hybrid approach, a `"toolcall"` event is dispatched on every incoming tool call _before_ executing the tool's `execute` function. The event handler can handle the tool call by calling the event's `preventDefault()` method, and then responding to the agent with `respondWith()` as shown above. If the event handler does not call `preventDefault()` then the browser's default behavior for tool calls will occur. The `execute` function for the requested tool is called. If a tool with the requested name does not exist, then the browser responds to the agent with an error.
```

bokand: I filed #8 as I'm not convinced we need the declarative form at all.

If we do decide we want this, an alternative approach is to include the declarative part only as a listing for informational/indexing purposes but keep only the procedural API for actually calling the functions. That is, an agent will still need to load the page to use the tool even with the declarative form so force registration and calling to happen in the same way.

We could avoid having a separate, differently shaped, API between the two forms by always requiring `agent.provideContext` even for declaratively provided tools. If we're worried about duplication between the declarative/procedural forms we could keep a parsed version of the declarative registration available as an object, e.g.

```json
// manifest.json: 
{
    "tools": [
        {
            "name": "add-todo",
            "description": "Add a new todo item to the list",
            ...
        }
    ]
}
```

```js
// js
window.agent.provideContext({
  tools: [
    window.agent.manifestTools['add-todo'],
  ]
});
```

docs/proposal.md
```diff
@@ -0,0 +1,262 @@
...
+
+```js
+async execute({ name, description, year, imageUrl }) {
+    addStamp(name, description, year, imageUrl);
+
+    return {
+        content: [
+            {
+                type: "text",
+                text: `Stamp "${name}" added successfully! The collection now contains ${stamps.length} stamps.`,
```
bokand, line 226: We'll need to specify how this structured output looks. WDYT about including something like the [recently added](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#tool )`outputSchema` in MCP?

No need to decide here but filed #9 to discuss.

docs/proposal.md
```diff
@@ -0,0 +1,262 @@
...
+Now the author needs to implement the tool. The tool needs to update the stamp database, and refresh the UI to reflect the change to the database. Since the code to do this is already available in the `addStamp()` function written earlier, the tool implementation is very simple and just needs to call this helper when an "add-stamp" tool call is received. After calling the helper, the tool needs to signal completion and should also provide some sort of feedback to the client application that requested the tool call. It returns a text message indicating the stamp was added:
+
+```js
+async execute({ name, description, year, imageUrl }) {
+    addStamp(name, description, year, imageUrl);
+
+    return {
+        content: [
+            {
+                type: "text",
```
bokand, line 225: Have y'all considered non text output? Thinking through examples this seems like it'd be very useful but I'm not sure yet how it'd look.

docs/proposal.md
```diff
@@ -0,0 +1,262 @@
...
+            },
+        ]
+    };
+}
+```
+### Future improvements to this example
+
+#### Use a worker
+
+To improve the user experience and make it possible for the stamp application to handle a large number of tool calls without tying up the document's main thread, the web developer may choose to move the tool handling into a dedicated worker script. Handling tool calls in a worker keeps the UI responsive, and makes it possible to handle potentially long-running operations. For example, if the user asks an AI agent to add a list of hundreds of stamps from an external source such as a spreadsheet, this will result in hundreds of tool calls.
```
bokand, line 236: Is this something that could happen at the application level? Tools are already async so an author could just postMessage the real work over to the worker already. Registering tools directly in a worker seems like maybe a small ergonomics improvement for that use case.

One interesting but maybe scary idea was tools in a service worker which could allow tool usage without a browsing context.

bokand (approved):

docs/proposal.md
bwalderman: Agree that we need a deeper discussion around the API. Keeping this doc as is for this PR so we can address this later. At the moment, proposal.md is just a temporary home for stuff that was moved out of explainer.md. I fully expect the API will change dramatically from what's written here, especially since the direction of prior art and the Script Tools API is leaning closer to something like a single `defineTool` call per tool.

docs/proposal.md
bwalderman: That's not something we considered while writing our original proposal, but worth investigating further.

docs/proposal.md
bwalderman: I was thinking each content item could be either a string or a Blob URI for images and audio. A Blob URI with a supported mimeType is returned to the browser, and that's exposed to the agent in a format that the agent expects (i.e. base64 embedded in JSON).

docs/proposal.md
jasonjmcghee:
```suggestion
Handling tool calls in the main thread with the option of delegating to workers serves a few purposes:
```

docs/proposal.md
bwalderman: Requiring tool registration to happen only in a top-level browsing context makes it easy to reason about which agent sees the tools (its the agent in the sidebar next to the tab). It also avoids potentially conflicting tool registrations from other frames and workers, and lets the browser enforce that the script registering the tools comes from the same origin. That was the main reason for not considering tool registration directly in workers.

An alternative to consider, we could define an object which holds the tools (like the `AutomationDelegate` from the Script Tools proposal). The object can only be created by a document in a top-level frame, but it can be transfered to workers via postMessage. So, for web devs who knows they want all of their tool handling to happen in a worker, they can just transfer the object to the worker and register the tools there. This allows the better ergonomics but still enforces the idea of one set of tools per tab.

PR #14, opened 2025-08-22
Move explainer content to README.md

bwalderman:
Moving the main explainer content to increase visibility.

I've left proposal.md in a subfolder for now. Eventually we should consider merging the API proposal into the main doc as well, but this should wait until we have a little more alignment across contributors on the actual API shape.

Added acknowledgements.


docs/explainer.md
anssiko, line 1: To avoid 404, I'd reduce this to a link to the new location.

README.md
```diff
@@ -1,13 +1,620 @@
 # WebMCP 🧪
 
-This is a repository for the [Web Machine Learning Community Group](https://www.w3.org/groups/cg/webmachinelearning/) ([join](https://webmachinelearning.github.io/community/#join)) to continue discussion on the [proposal](https://github.com/webmachinelearning/proposals/issues/12).
+_Enabling web apps to provide JavaScript-based tools that can be accessed by AI agents and assistive technologies to create collaborative, human-in-the-loop workflows._
 
-## Scope of Work
+> August 13, 2025
```
anssiko: It would be good to explain what this date means. Not introduced by this move PR, so to be handled separately. Explainers are typically not dated.

Issue #8, opened 2025-08-12
Should tools be a means for capability discovery?


bokand:
IIUC the primary motivation for the declarative form of tool registration is to enable discovery (e.g. by indexing in a search engine). Filing this issue for a broader discussion as to whether tools should be the means for discovery for an agent. We also initially went down this road but I think I convinced myself that tools shouldn't be the primitive for capability discovery but I'm curious to hear other perspectives.  My reasoning:

1. Tool != capability. Tools are meant as an _alternative_ to human UI so the capability should exist on the site even without the tool. e.g. For very simple apps/sites, if agents work perfectly well using the human UI, do we want to force authors to provide tools just to enable discovery?

2. Capabilities are a higher level abstraction than tools. An image editor might have tools like "Make Selection", "Select Brush", "Undo", "Redo", etc. These are all low level actions that only make sense in the broader context of the app. The app's _capability_ is "Image Editor" which would be better captured by some other mechanism.

3. Tools will become SEO'd. Authors will have an incentive to add spurious tools to their pages in order to surface themselves to agents, regardless of whether the add value.

4. Authors will now have a second axis along which to optimize tool descriptions for: agent usage and search engines. Given the difficult challenge of the former this seems like it will compromise agent quality

Thoughts?

mikechao:
> 1. For very simple apps/sites, if agents work perfectly well using the human UI, do we want to force authors to provide tools just to enable discovery?

I agree that for simple apps/sites current agents can work perfectly well using the human UI. But exposing tools to agents provide an additional benefits including faster speed of execution and cheaper costs.

Currently most(all?) web browsing agents use the tactic of take a screenshot, feed it to the LLM to figure out where the buttons are and click the button or use something like playwright. With the tools exposed to the agent, the LLM already know what actions it can take when the user asks it a question.

So for the [MCP-B Chrome Extension](https://github.com/MiguelsPizza/WebMCP) I created this [demo webpage](https://mcp-b-demo.pages.dev/) on the page we expose the same UI controls and MCP Tools.


Here is the demo page and the extension in action. We prompt the agent/Chrome extension with the following

```text
Can you do all of the following tasks?
1. add sleep to my todo
2. set the current project to mcp-b.ai
3. update the mood this is bananas
4. record my thought of "this is cool"
5. Increment the count by 42
```

https://github.com/user-attachments/assets/f699a9a7-577f-4821-84e0-e65e74178bc0

[...]

PR #19, opened 2025-08-28
Add new explainer for service workers

bwalderman:
Adding a supplemental explainer document for WebMCP within service workers. This would enable web developers to handle tool calls in service workers that can be activated on-demand, even if the web site isn't currently opened. This makes it possible to implement tools that work "in the background", but with the option to open windows for user input if needed.

### DIFF
--- a/docs/service-workers.md
+++ b/docs/service-workers.md
@@ -1,0 +1,279 @@
+# WebMCP for Service Workers
+
+> First published August 28, 2025
+>
+> Brandon Walderman <code>&lt;brwalder@microsoft.com&gt;</code><br>
+
+This is a supplemental explainer for the main WebMCP explainer at https://github.com/webmachinelearning/webmcp.
+
+## The Problem
+
+The initial explainer for [WebMCP](https://github.com/webmachinelearning/webmcp) covers how web pages expose context to AI agents that are already browsing a page. WebMCP's design makes it possible for in-browser AI integrations such as sidebar assistants to access context on currently opened browser tabs, and to operate the pages within those tabs.
+
+Sometimes, an agent may require context and tools from a site that the user doesn't currently have open. Perhaps the user is currently viewing a potential campsite on their favorite maps app and would like the agent to attempt to make a reservation while they continue to explore the map for potential nearby hikes. Ideally, the agent should be able to send a reservation request without navigating the tab or opening a new tab, either of which can be distracting to the user. What's needed is a way for the agent to discover an app that can handle the reservation request and a way to send that request in the background without necessarily showing any UI.
+
+## TL;DR
+
+This explainer describes a means for web sites to register service workers as WebMCP providers. Tool calls are handled in the service worker script where the web developer can handle the call directly, without needing to open any browser windows. If a tool requires user interaction though, the web developer can have their tool's execute function open a browser window and communicate with it via postMessage; For example, to let the user take care of sensitive steps like payment, or to present purchase confirmation.
+
+## Use Cases
+
+### Example – Productivity
+
+_Sarah is working on her laptop and needs to quickly capture some tasks without interrupting her current workflow. She's not currently on her todo list app, but the todo list app has previously registered WebMCP tools with the browser that the agent can use._
+
+**Sarah**: Add "Pick up dry cleaning" and "Call dentist about appointment" to my todo list.
+
+_The agent recognizes that `http://todoapp.example` has registered tools via its service worker. The service worker has exposed the following tool:_
+
+```javascript
+/**
+ * Adds a single item to the user's todo list.
+ * 
+ * item - A string representing the task to add to the list
+ * priority - Optional priority level: "low", "medium", "high". Defaults to "medium"  
+ * due_date - Optional due date in ISO format (YYYY-MM-DD)
+ */
+addTodoItem(item, priority, due_date)
+```
+
+_The agent calls `addTodoItem("Pick up dry cleaning", "medium")` and `addTodoItem("Call dentist about appointment", "medium")`. The service worker receives these calls and:_
+
+1.	Updates the user's locally stored todo list data
+2.	Syncs the changes with the app's backend API if network is available
+3.	Shows system notifications that the items were added
+
+**Agent**: Done! I've added both tasks to your todo list.
+
+_Sarah continues her work uninterrupted. Later, when she opens her todo app in a browser tab, the new items are already there, having been synced in the background._
+
+This demonstrates how service workers can handle tool calls entirely in the background, providing immediate value without requiring UI navigation or interrupting the user's current context. The service worker acts as an always-available local MCP service that can manage app state and provide simple user feedback through system notifications.
+
+### Example – Grocery Shopping
+
+_Mike is working from home and realizes he needs to order groceries for delivery. He wants to delegate meal planning and shopping to his agent but will need to review and complete the purchase manually._
+
+**Mike**: Plan this week's meals for me and order all the needed ingredients.
+
+_The agent recognizes that `http://freshmart.example` has registered tools via its service worker. The service worker has exposed several tools:_
+
+```javascript
+/**
+ * Searches for available products matching the given item name.
+ * 
+ * item_name - A string describing the product to search for
+ * returns - Array of products with id, name, price, and availability
+ */
+searchProducts(item_name)
+
+/**
+ * Adds a specific product to the user's shopping cart.
+ * 
+ * product_id - The ID of the product to add to cart
+ * quantity - Number of items to add (defaults to 1)
+ */
+addToCart(product_id, quantity)
+
+/**
+ * Initiates the order placement process. Opens a secure payment window
+ * if payment information is needed from the user.
+ * 
+ * returns - Completes the order or opens a payment interface as needed
+ */
+placeOrder()
+```
+
+_The agent performs the following sequence:_
+
+1.	Plans a week's worth of meals based on Mike's preferences (stored from previous interactions)
+2.	Calls `searchProducts("chicken breast")`, `searchProducts("rice")`, `searchProducts("broccoli")`, etc. for each ingredient needed
+3.	Selects appropriate products from the results and calls `addToCart(product_id, 1)` for each item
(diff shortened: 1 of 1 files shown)


docs/service-workers.md
```diff
@@ -0,0 +1,279 @@
...
+
+_The agent performs the following sequence:_
+
+1.	Plans a week's worth of meals based on Mike's preferences (stored from previous interactions)
+2.	Calls `searchProducts("chicken breast")`, `searchProducts("rice")`, `searchProducts("broccoli")`, etc. for each ingredient needed
+3.	Selects appropriate products from the results and calls `addToCart(product_id, 1)` for each item
+4.	The service worker updates the cart via backend API calls
+5.	Finally calls `placeOrder()`
+
+_ The service worker handles the `placeOrder()` call by:_
```
anssiko:
```suggestion
_The service worker handles the `placeOrder()` call by:_
```

MiguelsPizza:
Thanks for the mention @anssiko!

I'm planning to contribute a lot more to this group going forward. I've actually left my job to start a company around this idea, which is why I haven't been active here. The open source extension is much more of a proof of concept and I plan to transition it to being a tool for testing and writing webMCP servers rather than a serious client. I'll come share all my learnings next week post launch. I also plan to repurpose my old prompt-api extension as an webMCP polyfill and client and contribute it to the working group if you all are interested.

Here is a video of the service worker approach the MCP-B extension takes:

https://github.com/user-attachments/assets/32636706-3f70-4e9f-be78-f5ac1843246d


For context, the MCP-B extension aggregates WebMCP servers from all open tabs through the [mcpHub](https://github.com/MiguelsPizza/WebMCP/blob/main/apps/extension/entrypoints/background/src/services/mcpHub.ts) in the extension's service worker. It's basically acting as an MCP host like Claude Desktop, but for all the open tabs which have webMCP servers. It exposes the aggregate tools over a custom chrome ports transport which other MCP clients within the extension can connect to.


In terms of cross-origin, when executing tools, the extension service worker navigates to or focuses on the tab that holds the tool, then routes the call through the content script to the page's MCP server. I don't really do remote/background MCP execution. If we're going to be calling remote tools, we might as well just use a remote MCP server. I think the benefit of WebMCP servers a lot of times is that the human is in the loop and you can see tools executing. There's no guesswork about what's happening, and the human can stop anytime to prevent the agent from doing an action it shouldn't. Not saying remotely executing WebMCP tools doesn't have its place though.

Giving the agent access to the extension tab APIs solves many of the problems with discoverability. The agent generally has enough world knowledge to know that mail.google.com is the place to go to find and interact with email webMCP tools, etc.
On the security side, we've implemented an explicit consent model. When you visit a website with tools for the first time, you have to explicitly approve both the domain and the specific tools before they can be used. We are experimenting with hash-based verification where if any part of the tool changes (params, schema, description) it has to be re-approved. There's still the risk that a website owner could change what a tool actually does under the hood while keeping the signature the same, but at some point you're trusting the website just like you trust any web app you use. The fact that WebMCP tools can be disabled before starting an agentic workflow puts WebMCP miles above existing browser automation tools in terms of security, though there are still risks inherent to the agents themselves.

For data leakage between origins, I've been thinking about a clipboard-like approach where website tools can request to write to a client-managed clipboard without the data ever entering the agent's context. So if you had sensitive info like a social security number on one site, the agent could store it in this clipboard and then when it needs to use it elsewhere, it would need explicit user permission to paste it, but the agent never actually sees the data. This could help maintain separation between origins while still enabling workflows.

The multi-origin access is definitely powerful but I can see why you're considering single-origin limitation as a starting point. Your service worker approach would actually solve the tab-switching friction nicely for single-origin workflows where you do want that background execution without the visual context switching, and this makes sense for WebMCP clients which live in a webapp rather than a browser extension.

docs/service-workers.md
```diff
@@ -0,0 +1,279 @@
...
+
+This is a supplemental explainer for the main WebMCP explainer at https://github.com/webmachinelearning/webmcp.
+
+## The Problem
+
+The initial explainer for [WebMCP](https://github.com/webmachinelearning/webmcp) covers how web pages expose context to AI agents that are already browsing a page. WebMCP's design makes it possible for in-browser AI integrations such as sidebar assistants to access context on currently opened browser tabs, and to operate the pages within those tabs.
+
+Sometimes, an agent may require context and tools from a site that the user doesn't currently have open. Perhaps the user is currently viewing a potential campsite on their favorite maps app and would like the agent to attempt to make a reservation while they continue to explore the map for potential nearby hikes. Ideally, the agent should be able to send a reservation request without navigating the tab or opening a new tab, either of which can be distracting to the user. What's needed is a way for the agent to discover an app that can handle the reservation request and a way to send that request in the background without necessarily showing any UI.
```
khushalsagar, line 13: > Ideally, the agent should be able to send a reservation request without navigating the tab or opening a new tab, either of which can be distracting to the user.

I'm not seeing why loading a tab in the background (if we don't want to take the user out of the current context) is not enough. The core sell of WebMCP is journeys where the user needs to interact with the service frequently enough that there's benefit to having the site's UI on the client. If something can be accomplished by the Agent largely interacting with the site in the background then wouldn't remote MCP be better? I'm echoing @MiguelsPizza's take here: "If we're going to be calling remote tools, we might as well just use a remote MCP server. I think the benefit of WebMCP servers a lot of times is that the human is in the loop and you can see tools executing."

> What's needed is a way for the agent to discover an app that can handle the reservation request

The discovery problem is orthogonal to whether tool execution requires a tab or is backgrounded. My thoughts there align with @bokand's feedback on https://github.com/webmachinelearning/webmcp/issues/8.

docs/service-workers.md
```diff
@@ -0,0 +1,279 @@
...
+
+#### HTTP endpoints handled via service worker fetch event
+
+Some existing web app manifest features like Share Targets, protocol handlers, and shortcuts expose app actions to the host system as URLs that can be fetched via HTTP GET or POST. These can be intercepted and handled in a service worker by listening to fetch events or passed through to the server-side by default.
+
+In the case of a feature like share targets, the action may be handled in script for the best user experience, but there is a reasonable default behavior of falling back to a simple HTTP request to a server, which is why framing share targets as HTTP endpoints in a declarative app manifest makes sense.
+
+WebMCP is intended to align closely with MCP architecture so that developers that have experience with one can easily apply their skills to the other. In MCP, tools are not individual HTTP endpoints. Handling WebMCP tools with dedicated callback functions in script instead of as HTTP fetch events aligns with how tools are treated in backend MCP frameworks.
+
+## Security Considerations
```
khushalsagar, line 250: These considerations are the same irrespective of foreground tab interaction or service worker based execution, right?

docs/service-workers.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Appendix A: Discovery
+
+Before an agent can interact with the web through tool calling, it needs to discover a web site to interact with. In the case of the WebMCP API as currently designed, this problem is implicitly solved because the user must first navigate to the page. For an agent to access the tools from a web site that the user isn't currently browsing, it needs some means of discovering that site and getting information about the tools and context that site provides.
+
+Discovery is a complex topic which goes beyond the scope of this explainer but merits some discussion here for the purpose of understanding the feature end-to-end. Fundamentally, tool discovery requires some way for agents to obtain the URL of a site that has tools relevant to the user's tasks, and a way to install that site's service worker for handling tool requests. There are many potential mechanisms for this, including but not limited to:
```
khushalsagar, line 262: > Fundamentally, tool discovery requires some way for agents to obtain the URL of a site that has tools relevant to the user's tasks

Are you considering UI actuation in this design or is the Agent limited to sites which support WebMCP?

bwalderman:
@MiguelsPizza 

> If we're going to be calling remote tools, we might as well just use a remote MCP server. 

That's always an option, but one advantage of using WebMCP for background tasks is that if the user is already logged in to the site/app then the WebMCP service worker will have those credentials. Otherwise, the service worker can open a tab for the user to authenticate.

As I understand it, auth with remote MCP servers is still quite rough, and requires creating and managing auth tokens.

MiguelsPizza:
@bwalderman That's a fair point, it would be beneficial for readonly tasks in particular.

 What are your thoughts on supporting [elicitation](https://modelcontextprotocol.io/specification/draft/client/elicitation) in the webMCP spec. It would be a good way to focus on the remote tab in the case user input is needed to unblock the model.

docs/service-workers.md
bwalderman: Yes, these considerations apply to both.

docs/service-workers.md
bwalderman: This design is mainly talking about discovery for sites that support service worker WebMCP.

Discovering a website that can help the user with X tasks through actuation or on-page WebMCP tools is accomplished with a web search and a navigation. 

For service-worker based tools though, we want to support the caser where the user never actually navigates to the web site (they agent discovers and registers the service worker directly). So, this design discusses what infrastructure an agent would need to find and activate that service worker.

bwalderman:
I'm going to open a separate issue to discuss elicitation to avoid this PR thread going too off topic :)

docs/service-workers.md
khushalsagar: I think we need to reconcile the broader question of how we see these tools as means for discovering capabilities of sites. https://github.com/webmachinelearning/webmcp/issues/8 is focused on that. While these tools can be one way of understanding what a site can accomplish, indexing using content built for human consumption in the site will be the better source of truth of a site's capabilities for a while.

That said, I still don't see how the discovery aspect is different between service worker vs on-page WebMCP tools. The Agent is making 2 decisions when the user makes a query:

1. Which site/sites should I load for this query.
2. Should I interact with this site in the foreground or background.

Discovery will happen first for 1). Once the site has been decided, whether it's loaded in a tab or there's just-in-time fetching of service worker (as is proposed here) will be a separate decision from discovery.

You could say that first an Agent decides whether the task should be executed in the background. If yes, then it biases towards sites which support WebMCP in a service worker. But I don't see a world where that's optimal for the user. The Agent should be selecting the site based on it's capabilities (and using the best integration option available), not based on the type of integration it supports.

Issue #25, opened 2025-09-16
Core design principles for WebMCP


khushalsagar:
We've made some implicit assumptions in how much WebMCP shares in the design for the underlying MCP protocol. There's good reasons for these but it seems worth outlining explicitly and ensuring there's consensus.

MCP defines 2 layers for how server-client communication is setup. There's a [data layer](https://modelcontextprotocol.io/docs/learn/architecture#data-layer-2) which is the request/response definition for initialization, listing and executing tools etc: what the control messages are. And the [transport layer](https://modelcontextprotocol.io/docs/learn/architecture#transport-layer) underneath the data layer which manages how these messages are exchanged. The data layer protocol further has core [primitives](https://modelcontextprotocol.io/docs/learn/architecture#primitives) that an MCP server can expose: `tools`, `resources` and `prompts`.

WebMCP's current proposal uses `tools` from MCP as-is (other primitives can also be considered) but not the data layer. This implies that there will need to be a translation layer for an MCP compatible client to use a WebMCP based server. Let's call this the "SDK option".

The other option is to use the data layer from MCP as-is and have the browser provide the transport layer. The browser manages the routing of messages between an MCP client and the site (could be an event handler) but the messages themselves are opaque to the browser.  Let's call this the "Transport Only option".

I'm in favour of the SDK option for the following reasons. Credits to @bokand for the exploration of these options:

1. Doesn’t directly couple the Web to a specific MCP version. The fact that the control flow is intermediated by the browser, instead of being opaque messages helps in the following ways:
    - The browser can ensure backwards compatibility as the MCP protocol evolves.
    - The browser can impose security principles which don't make sense for MCP. For instance, an embedder controling whether cross-origin iframes can use WebMCP or a new capability in the MCP API.
2. There will be interaction patterns where WebMCP and MCP fundamentally don't align. For instance, MCP *needs* the Agent/MCP client to mediate all interaction with the user but WebMCP doesn't (see https://github.com/webmachinelearning/webmcp/issues/21).
3. Easier to ensure API ergonomics align with the web. For instance, tool response can use `img` or `video` elements for multi-modal output.
4. Declarative APIs as brought up in https://github.com/webmachinelearning/webmcp/issues/22 are easier to add.

I think 2) could be added to MCP, a concept of "MCP services on the user's device" but the rest of the reasons will still be valid. So we could establish the following best practices for any new feature request:

- Explore if/how this problem is solved in MCP.
[...]

PR #36, opened 2025-10-09
Add explaination for intersection with MCP

khushalsagar:
Documenting the resolution at https://github.com/webmachinelearning/webmcp/issues/25.


khushalsagar:
Do let me know if there's a better place to add this since the question has come up a few times.

Issue #41, opened 2025-10-17
Support image input/output in tools


khushalsagar:
The API proposal doesn't include support for image input to the tool and for the tool output to include images. This seems like a basic use-case we should support. Worth looking at how MCP is doing it but we'll need something web specific, allowing devs to use image elements for this as an example.

MiguelsPizza:
Providing some reference material for the tool result types in the MCP spec:

Included in the [ToolResult object ](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts#L778-L804)is an array of [Content Blocks](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts#L1079-L1084)

```typescript
export type ContentBlock =
  | TextContent
  | ImageContent
  | AudioContent
  | ResourceLink
  | EmbeddedResource;

export interface CallToolResult extends Result {
  /**
   * A list of content objects that represent the unstructured result of the tool call.
   */
  content: ContentBlock[];
...
```

Images can be returned as an [ImageContent](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts#L1108-L1135) or an [EmbeddedResource](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts#L725-L744). Both of which request the image be represented as a Base64-encoded string.

```typescript
/**
 * An image provided to or from an LLM.
 */
export interface ImageContent {
  type: "image";

  /**
   * The base64-encoded image data.
   *
   * @format byte
   */
  data: string;

  /**
   * The MIME type of the image. Different providers may support different image types.
   */
  mimeType: string;

  /**
   * Optional annotations for the client.
   */
  annotations?: Annotations;

  /**
   * See [General fields: `_meta`](/specification/2025-06-18/basic/index#meta) for notes on `_meta` usage.
   */
  _meta?: { [key: string]: unknown };
}
```

I believe most inference providers accept base64. It would nice if the browser did the conversion so the server code could be as simple as this

```typescript
window.navigator.modelContext.registerTool({
    name: "get_product_image",
    description: "Get the product image from the page",
    inputSchema: {
      type: "object",
      properties: {
        productId: { type: "string" }
      }
    },
    handler: async ({ productId }) => {
      const img = document.querySelector(`#product-${productId} img`);
      
[...]

PR #42, opened 2025-10-21
Minor formatting corrections

rviscomi:
Updated README to fix formatting issues and broken links

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,7 +305,7 @@ _The agent calls `getDresses(6)` and receives a JSON object:_
 ```
 
 > [!Note]
-> How to pass images and other non-textual data is something we should improve (See [Issue #10](https://github.com/webmachinelearning/webmcp/issues/10))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -285,27 +285,26 @@ _The agent calls `getDresses(6)` and receives a JSON object:_
...
         },
         ...
     ]
 }
 ```
 
 > [!Note]
-> How to pass images and other non-textual data is something we should improve 
-> Issue #10
+> How to pass images and other non-textual data is something we should improve (See [Issue #10](https://github.com/webmachinelearning/webmcp/issues/10))
```
rviscomi: Double checking: is this the correct issue?

There's also https://github.com/webmachinelearning/webmcp/issues/41

anssiko: Yes I think so. While here, do you want to resolve the conflict?

rviscomi: Done!

anssiko: I stand corrected. #41 seems like a better fit.

rviscomi: Ok, updated

anssiko (approved):

PR #59, opened 2025-12-08
Add section on tool implementation as attack targets

victorhuangwq:


docs/security-privacy-considerations.md
```diff
@@ -151,6 +151,44 @@ navigator.modelContext.registerTool({
...
+##### 3. Tool Implementation as Attack Targets
+
+Websites exposing valuable functionality through WebMCP tools can themselves become targets for attacks.
+
+- **Threat Actor**: Malicious users controlling or manipulating agents using WebMCP
+- **Target**: Websites implementing valuable or sensitive WebMCP tools
+- **Assets at Risk**:
+  - High-value actions exposed by the tool (e.g., database access, transactions)
+
+**How It Works**: Websites have high-value functionality (e.g., password resets, transactions) through their UI. Agents capable of manipulating rendered elements can already interact with this functionality. When websites additionally expose such functionality via WebMCP tools, they create another potential target for malicious agents.
```
victorhuangwq, line 163: @khushalsagar I'm opening a new PR to add back the 3rd point for prompt injection attacks. I'm focusing here more on the fact that the webmcp tools themselves are potentially a target and clearing them up here.

khushalsagar: The threat actor listed here is a malicious user but that doesn't make sense to me. The user already has access to this functionality via the site.

I can imagine a malicious origin manipulating an agent to use a high-value action offered by another site. But that is covered by the attack vectors above. So not following what additional threats this is meant to cover...

victorhuangwq: I must have glossed over the threat actor when I changed the focus from prompt injection on the input, to the new point on tool implementation as attack targets. 

Let me rephrase it - I think this is more of any malicious actors that is able to gain control of agents that has access to WebMCP tool

khushalsagar: Hmmm, I'm still not completely convinced that this is highlighting something distinct from other attack vectors. The site has to trust that the Agent being provided with these tools has appropriate mitigations in place so an attacker can't take control of it. Especially with browser-agent which is conceptually trusted the same way the user-agent/browser is.

But doesn't hurt to have this. We can discuss on an issue and conclude if anything is needed for this.

victorhuangwq:
Could I get help merging this? thank you!

PR #61, opened 2026-01-06
Fix Javascript examples in proposal

beaufortfrancois:
I've noticed the following Javascript issues in the proposal examples:
- `execute: async execute(...) => {` is not valid. It should be `execute: (...) => {`.
- `buyProduct(input` should be `buyProduct({ product_id }` as we don't use `input`.
- `"modelContext" in window` is wrong. It should be `"modelContext" in window.navigator`.
- Remove extra `async` that are not needed.

### DIFF
--- a/docs/proposal.md
+++ b/docs/proposal.md
@@ -68,7 +68,7 @@ window.navigator.modelContext.provideContext({
                 },
                 required: ["text"]
             },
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
@khushalsagar Please review

beaufortfrancois:
gentle ping

Issue #9, opened 2025-08-12
Should output also have a schema?


bokand:
The explainer includes an example returning some structured output:

```
return {
        content: [
            {
                type: "text",
                text: `Stamp "${name}" added successfully! The collection now contains ${stamps.length} stamps.`, 
            },
        ]
    };
```

Should we include an `outputSchema` as [MCP recently added](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#tool)?

I'm not sure the schema itself helps (the return is already a JavaScript object...) but descriptions might help the agent understand what the output means?

bwalderman:
The motivation for `outputSchema` in MCP is discussed here https://github.com/modelcontextprotocol/modelcontextprotocol/pull/356

Seems like if it makes sense for classic MCP, it would make sense for WebMCP too.

khushalsagar:
This will need to be supported for the declarative API as well. The case where JS preventDefaults the submit event, the result is similar to the imperative tools case. So it could just be an attribute on the `form` element.

The cross-document case where we're consider `ld-json` for the result needs more thought. @domfarolino FYI.

anssiko:
> [RESOLUTION](https://www.w3.org/2026/02/05-webmachinelearning-minutes.html#ed89): Add outputSchema to the imperative API and investigate how to associate outputSchema with declarative WebMCP (issue [#9](https://github.com/webmachinelearning/webmcp/issues/9))

PR #75, opened 2026-02-05
Add WebIDL and descriptions to spec draft

bwalderman:
An initial pass at WebIDL for the current proposed API, and short descriptions of each property/method.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/75.html" title="Last updated on Feb 11, 2026, 9:31 PM UTC (98b1ba3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/75/78ae4d7...98b1ba3.html" title="Last updated on Feb 11, 2026, 9:31 PM UTC (98b1ba3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -142,6 +142,11 @@ interface ModelContext {
     <p>Registers the provided context (tools) with the browser. This method clears any pre-existing tools and other context before registering the new ones.
   </dd>
 
+  <dt><code><var ignore>navigator</var>.{{Navigator/modelContext}}.{{ModelContext/clearContext()}}</code></dt>
+  <dd>
+    <p>Unregisters all context (tools) with the browser.
+  </dd>
+
   <dt><code><var ignore>navigator</var>.{{Navigator/modelContext}}.{{ModelContext/registerTool(tool)}}</code></dt>
   <dd>
     <p>Registers a single tool without clearing the existing set of tools. The method throws an error, if a tool with the same name already exists, or if the {{ModelContextTool/inputSchema}} is invalid.
@@ -153,22 +158,34 @@ interface ModelContext {
   </dd>
 </dl>
 
-<dl>
-  <dt><dfn method for="ModelContext">provideContext(options)</dfn></dt>
-  <dd>
-    <p>TODO: fill this out</p>
-  </dd>
+<div algorithm>
+The <dfn method for=ModelContext>provideContext(<var ignore>options</var>)</dfn> method steps are:
 
-  <dt><dfn method for="ModelContext">registerTool(tool)</dfn></dt>
-  <dd>
-    <p>TODO: fill this out</p>
-  </dd>
+1. TODO: fill this out.
 
-  <dt><dfn method for="ModelContext">unregisterTool(name)</dfn></dt>
-  <dd>
-    <p>TODO: fill this out</p>
-  </dd>
-</dl>
+</div>
+
+<div algorithm>
+The <dfn method for=ModelContext>clearContext()</dfn> method steps are:
+
+1. TODO: fill this out.
+
+</div>
+
+
+<div algorithm>
+The <dfn method for=ModelContext>registerTool(<var ignore>tool</var>)</dfn> method steps are:
+
+1. TODO: fill this out.
+
+</div>
+
+<div algorithm>
+The <dfn method for=ModelContext>unregisterTool(<var ignore>name</var>)</dfn> method steps are:
+
+1. TODO: fill this out.
+
+</div>
 
 <h4 id="model-context-options">ModelContextOptions Dictionary</h4>
 
@@ -182,13 +199,6 @@ dictionary ModelContextOptions {
   <dt><code><var ignore>options</var>["{{ModelContextOptions/tools}}"]</code></dt>
   <dd>
     <p>A list of {{ModelContextOptions/tools}} to register with the browser. Each tool name in the list is expected to be unique.
-  </dd>
-</dl>
-
-<dl>
-  <dt><dfn dict-member for="ModelContextOptions">tools</dfn></dt>
-  <dd>
-    <p>TODO: fill this out</p>
   </dd>
 </dl>
 
@@ -241,46 +251,12 @@ callback ToolExecuteCallback = Promise<any> (object input, ModelContextClient cl
   </dd>
 </dl>
 
-<dl>
-  <dt><dfn dict-member for="ModelContextTool">name</dfn></dt>
-  <dd>
-    <p>TODO: fill this out</p>
-  </dd>
-
-  <dt><dfn dict-member for="ModelContextTool">description</dfn></dt>
-  <dd>
-    <p>TODO: fill this out</p>
-  </dd>
-
-  <dt><dfn dict-member for="ModelContextTool">inputSchema</dfn></dt>
-  <dd>
-    <p>TODO: fill this out</p>
-  </dd>
-
-  <dt><dfn dict-member for="ModelContextTool">execute</dfn></dt>
-  <dd>
-    <p>TODO: fill this out</p>
-  </dd>
-
-  <dt><dfn dict-member for="ModelContextTool">annotations</dfn></dt>
-  <dd>
-    <p>TODO: fill this out</p>
-  </dd>
-</dl>
-
 The {{ToolAnnotations}} dictionary provides optional metadata about a tool:
 
 <dl class="domintro">
   <dt><code><var ignore>annotations</var>["{{ToolAnnotations/readOnlyHint}}"]</code></dt>
   <dd>
     <p>If true, indicates that the tool does not modify any state and only reads data. This hint can help [=agents=] make decisions about when it is safe to call the tool.
-  </dd>
-</dl>
-
-<dl>
-  <dt><dfn dict-member for="ToolAnnotations">readOnlyHint</dfn></dt>
-  <dd>
-    <p>TODO: fill this out</p>
   </dd>
 </dl>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+</xmp>
+
+<h3 id="model-context-container">ModelContext Interface</h3>
+
+The {{ModelContext}} interface provides methods for web applications to register and manage tools that can be invoked by [=agents=].
+
+<xmp class="idl">
+[Exposed=Window, SecureContext]
+interface ModelContext {
+  undefined provideContext(optional ModelContextOptions options = {});
```
beaufortfrancois, line 132: Chrome currently has `undefined clearContext();`. Does that mean calling `provideContext()` with no parameters clears context or should we call it with `null`?

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+
+<dl>
+  <dt><dfn method for="ModelContext">provideContext(options)</dfn></dt>
+  <dd>
+    Registers the provided context (tools) with the browser. This method clears any pre-existing tools and other context before registering the new ones.
+  </dd>
+
+  <dt><dfn method for="ModelContext">registerTool(tool)</dfn></dt>
+  <dd>
+    Registers a single tool without clearing the existing set of tools. If a tool with the same name already exists, the method throws an Error.
```
beaufortfrancois: It will also raise an error in the following cases:
- tool name is missing 
- description is missing
- input schema is invalid

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+
+<h3 id="model-context-tool">ModelContextTool Dictionary</h3>
+
+The {{ModelContextTool}} dictionary describes a tool that can be invoked by [=agents=].
+
+<xmp class="idl">
+dictionary ModelContextTool {
+  required DOMString name;
+  required DOMString description;
+  required object inputSchema;
```
beaufortfrancois: Does this need to be required? If it's not there, it could default to `{ type: 'object', properties: {} }`. WDYT?

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+  ToolExecuteCallback execute;
```
beaufortfrancois: I think the execute function should be mandatory though no?

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+  ToolAnnotations annotations;
+};
+
+dictionary ToolAnnotations {
+  boolean readOnly;
```
beaufortfrancois: Why not naming it `readOnlyHint` as in MCP?

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
 };
 
+callback ToolExecuteCallback = ToolResponse (object input, ModelContextClient client);
```
beaufortfrancois: FYI Chromium currently defines this as `Promise<any>(any... parameters);`.

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+<h3 id="navigator-extension">Extensions to the Navigator Interface</h3>
+
+The {{Navigator}} interface is extended to provide access to the {{ModelContext}}.
+
 <xmp class="idl">
 partial interface Navigator {
-  [SecureContext, SameObject] readonly attribute ModelContextContainer modelContext;
+  [SecureContext, SameObject] readonly attribute ModelContext modelContext;
+};
+</xmp>
```
domfarolino, line 123: One small thing is that you'll need to define what happens when `modelContext` is called and how its backing value is initialized. You'll want something like this: https://wicg.github.io/fenced-frame/#window-fence.

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
 <h2 id="api">API</h2>
 
 <!--
-TODO: Convert API described in proposal.md into WebIDL, sketch initial algorithms, define attributes and methods etc.
+TODO: Sketch initial algorithms, define attributes and methods etc.
 https://github.com/webmachinelearning/webmcp/blob/main/docs/proposal.md#api
 https://dlaliberte.github.io/bikeshed-intro/#a-strategy-for-incremental-development
 -->
 
+<h3 id="navigator-extension">Extensions to the Navigator Interface</h3>
```
domfarolino:
```suggestion
<h3 id="navigator-extension">Extensions to the {{Navigator}} Interface</h3>
```

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+[Exposed=Window, SecureContext]
+interface ModelContext {
+  undefined provideContext(optional ModelContextOptions options = {});
+  undefined registerTool(ModelContextTool tool);
+  undefined unregisterTool(DOMString name);
+};
+</xmp>
+
+<dl>
+  <dt><dfn method for="ModelContext">provideContext(options)</dfn></dt>
```
domfarolino: Because this is just a descriptive list of what the methods do, these should not contain the method definitions themselves. I think this should basically be what we call a "dom intro" box (for web developers to read) like https://github.com/WICG/nav-speculation/blob/main/prerendering.bs#L450-L457, and then below we want the actual method definitions, like https://github.com/WICG/fenced-frame/blob/master/spec.bs#L2042-L2045 does.

For now the method definitions can just have `<p class="XXX">TODO: fill this out</p>` in them, to represent that work needs to be done to fill them out.

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+
+<h3 id="model-context-options">ModelContextOptions Dictionary</h3>
+
+<xmp class="idl">
+dictionary ModelContextOptions {
+  sequence<ModelContextTool> tools = [];
+};
+</xmp>
+
+<dl>
```
domfarolino: Likewise, these should be domintro boxes since they are descriptions of the requirements put upon each member, but the actual normative enforcement of this will appear in real algorithms that operate on the data and enforce constraints. The definitions for each member should be inside the IDL (they are by default), and I think these lines here should reference them as `{{ModelContextOptions/tools}}` and so on.

index.bs
domfarolino: +1 this should be mandatory

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+  <dt><dfn method for="ModelContextClient">requestUserInteraction(callback)</dfn></dt>
+  <dd>
+    Asynchronously requests user input during the execution of a tool.
+    The callback function is invoked to perform the user interaction (e.g., showing a confirmation dialog), and the promise resolves with the result of the callback.
+  </dd>
+</dl>
+
+<h3 id="tool-response">ToolResponse Dictionary</h3>
+
+The {{ToolResponse}} dictionary represents the structured content response returned by a tool execution.
```
domfarolino: Per the other comment in this CL, I'm not sure that this dictionary is needed, since the response should probably always be a Promise, unless there's been consensus to change this elsewhere?

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+    An array of content items representing the tool's response to the [=agent=].
+  </dd>
+</dl>
+
+The {{ToolContent}} dictionary represents an individual content item in a tool response:
+
+<dl>
+  <dt><dfn dict-member for="ToolContent">type</dfn></dt>
+  <dd>
+    The type of content. For example, "text" for plain text content.
```
domfarolino: I'm hopeful we don't need to rely on this, and instead we can just move forward with https://github.com/webmachinelearning/webmcp/issues/9 to provide all of the necessary output schema information to the model for tool execution.

index.bs
```diff
@@ -47,27 +47,191 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+    Registers a single tool without clearing the existing set of tools. If a tool with the same name already exists, the method throws an Error.
+  </dd>
+
+  <dt><dfn method for="ModelContext">unregisterTool(name)</dfn></dt>
+  <dd>
+    Removes the tool with the specified name from the registered set.
+  </dd>
+</dl>
+
+<h3 id="model-context-options">ModelContextOptions Dictionary</h3>
```
domfarolino: Since this and all dictionaries below are associated with the `ModelContext` interface, how about we make this `<h4>` so that they're nested under the description for that interface, since that's the only place they're used?

index.bs
bwalderman: Thanks. I'll mention an error being thrown for input schema invalid. I would expect a TypeError on missing tool name or description is implied by the `required` keyword for these properties on `ModelContextTool`.

index.bs
bwalderman: Planning to cover that in a follow up PR along with some initial method algorithms.

index.bs
bwalderman: I guess there isn't quite consensus on a structured response format yet [#41]. Removing this for now.

index.bs
bwalderman: Updated the return Promise to `any`. I'm keeping the parameters as-is since I think these are the current consensus, even if `client` (used for elicitation) isn't implemented in Chromium yet.

index.bs
```diff
@@ -47,27 +47,223 @@ https://github.com/webmachinelearning/webmcp/blob/main/docs/security-privacy-con
...
+<h3 id="model-context-container">ModelContext Interface</h3>
 
+The {{ModelContext}} interface provides methods for web applications to register and manage tools that can be invoked by [=agents=].
+
+<xmp class="idl">
 [Exposed=Window, SecureContext]
-interface ModelContextContainer {
+interface ModelContext {
+  undefined provideContext(optional ModelContextOptions options = {});
+  undefined clearContext();
```
beaufortfrancois, line 133: Provide a domintro for `clearContext()`.

domfarolino:
I rebased this PR on top of `main`, and pushed a commit resolving a few other comments.

Issue #22, opened 2025-09-05
Declarative API Equivalent


EisenbergEffect:
I'm curious as to why this is being pursued exclusively as a JavaScript API. It seems that most of this could be done entirely declaratively, or at least that there should be a declarative equivalent. Simply adding a few attributes, combined with existing HTML semantics should do the trick for a lot of scenarios.

```html
<form action="/todos" method="post" tool-name="add-todo" tool-description="Add a new todo item to the list">
  <input type="text" name="description" required tool-prop-description="The text of the todo item">
  <button type="submit">Add Todo</button>
</form>
```

This might even be a good opportunity to improve forms with features that folks have been asking for for decades, such as:

* Supporting more `action` values.
* Enable a form to use AJAX
* Do different things with the response from an AJAX form submission (e.g. pass it to the agent vs. put it somewhere in the HTML document).

While agent interaction is a "big thing" right now, it would be great to think of this in the context of the broader, long-term web platform, and come up with a consistent, extensible solution. We should also consider how content creators without JS experience will be able to participate in the agent ecosystem, and strive to be as inclusive as we can.

vitalets:
I also thought about this approach. But how to declare *listTodos* tool in this manner?

EisenbergEffect:
> I also that about this approach. But how to declare _listTodos_ tool in this manner?

I think that could be done with a simple link.

```html
<a href="/todos" tool-name="list-todos">All Todos</a>
```

Whenever an `a` or `form` tool is invoked by an agent, the browser could just use `fetch` to make the request and pass the response back to the agent that invoked the tool.

MiguelsPizza:
>Whenever an a or form tool is invoked by an agent, the browser could just use fetch to make the request and pass the response back to the agent that invoked the tool.

The counter argument here would be that this limits the control the tool creator has over things like param and response parsing. One of the strengths of the MCP spec is you can determine exactly what information to return to the model's context via filtering with existing javascript APIs. 

How would we filter the output of href="/todos" so it does not return irrelevant data to the agent's context? Would you build backend endpoints? One for the user and one for the agent?

EisenbergEffect:
> How would we filter the output of href="/todos" so it does not return irrelevant data to the agent's context? Would you build backend endpoints? One for the user and one for the agent?

This can all be handled with standard HTTP/REST patterns.

[...]

PR #76, opened 2026-02-05
Declarative API Explainer

domfarolino:
This is a follow-up of https://github.com/webmachinelearning/webmcp/pull/26 to address https://github.com/webmachinelearning/webmcp/issues/22. Note that #26 has been superseded by this PR, per https://github.com/webmachinelearning/webmcp/pull/26#issuecomment-3893501689.

### DIFF
--- a/declarative-api-explainer.md
+++ b/declarative-api-explainer.md
@@ -116,6 +116,11 @@ we've come up with should be supported by the community as a general approach.
 
 ### Getting the form response to the agent
 
+This topic is currently under debate; see https://github.com/webmachinelearning/webmcp/issues/135.
+
+<details>
+<summary>Click to read the `application/ld+json` proposal before the above issue was filed</summary>
+
 When a form element performs a navigation, the first `<script type=application/ld+json>` tag on the
 target page is used as the cross-document tool's "response" that gets sent to the model.
 
@@ -125,6 +130,7 @@ However, this is technically TBD at the moment.
 
 When the form element does *NOT* perform a navigation, JavaScript can hand-craft the response to the
 agent via the `SubmitEvent#respondWith()` method described below.
+</details>
 
 ### Pseudo-classes
 
@@ -168,7 +174,7 @@ interface SubmitEvent : Event {
 
 **`toolactivated` and `toolcanceled` events
 
-We introduce these events that get fired at the `Window` object when a WebMCP tool is run, and when
+We introduce these events that get fired at the `ModelContext` object when a WebMCP tool is run, and when
 its invocation is canceled.
 
 The `toolactivated` event gives the developer a hook to perform any actions, such as bringing the
@@ -176,6 +182,10 @@ form to the user's attention, once a declarative tool is filled out but before i
 (This presumes the absence of the `toolautosubmit` attribute). This event can be seen as the
 JavaScript equivalent of the [`:tool-form-active` pseudo-class](#pseudo-classes).
 
+When the agent cancels a tool call (perhaps because a user has instigated another turn of the
+conversation, obviating the need for the pending tool), the `toolcanceled` event is fired. Note that
+this event does not fire when the site itself cancels the tool, due to removing the form element or
+changing its name or description.
 
 Some open questions:
 


maerzhase:
The proposal looks great! its exactly what i have in mind to support as additional tools via [`ai11y`](https://github.com/maerzhase/ai11y).

domfarolino:
@bwalderman, @mfreed7, and @bvandersloot, could you take a look?

declarative-api-explainer.md
```diff
@@ -0,0 +1,162 @@
...
+[`min`](https://html.spec.whatwg.org/C#attr-input-min) is TBD. We need to concretely specify how
+various form-associated elements like `<input>` and `<select>` reduce to a JSON Schema that includes
+`anyOf`, `oneOf`, and `maximum`/`mininum` declarations.
+
+Chromium is implementing a loose version of this and will conduct testing/trials to see if what
+we've come up with should be supported by the community as a general approach.
+
+### Getting the form response to the agent
+
+TODO: Mention application/json-ld responses, and so on.
```
bvandersloot-mozilla: I'm curious for more info on how you envision this. If you mean dropping additional information into a `<script type="application/ld+json">` on the form navigation's result, I think that is a good direction.

declarative-api-explainer.md
```diff
@@ -0,0 +1,162 @@
...
+## Integration with other imperative API bits
+
+It's an open question as to whether [an
+`outputSchema`](https://github.com/webmachinelearning/webmcp/issues/9) makes sense for declarative
+WebMCP tools, and therefore if the `agentResponse` Promise passed to `SubmitEvent#respondWith()`
+must resolve to an object conforming to such schema.
+
+It is TBD how *declarative* WebMCP tools will be exposed to any interface that exposes a site's
+tools to JavaScript. See https://github.com/webmachinelearning/webmcp/issues/51 for context. Should
+a declarative WebMCP tool be able to be invoked from such an interface, should it exist in the
```
bvandersloot-mozilla, line 211: Probably? But I understand the caution.

declarative-api-explainer.md
```diff
@@ -0,0 +1,162 @@
...
+
+> [!WARNING]
+> For declarative, should they be fired at `Window` or at the `<form>` that registered the tool in
+> the first place, and bubble up to the document that way?
+
+## Integration with other imperative API bits
+
+It's an open question as to whether [an
+`outputSchema`](https://github.com/webmachinelearning/webmcp/issues/9) makes sense for declarative
+WebMCP tools, and therefore if the `agentResponse` Promise passed to `SubmitEvent#respondWith()`
```
bvandersloot-mozilla, line 206: This also relates to [the issue of who owns schema validation.](https://github.com/webmachinelearning/webmcp/issues/92)

Inferring the output schema is impossible? So we'd have to permit it in the form attribute. Not impossible, but potentially very bulky.

declarative-api-explainer.md
```diff
@@ -0,0 +1,162 @@
...
+  tooldescription="This form searches flights and displays [...]"
+  toolautosubmit>
+```
+
+The `toolname` attribute is analogous to the imperative API's
+[`ModelContextTool#name`](https://webmachinelearning.github.io/webmcp/#dom-modelcontexttool-name),
+while `tooldescription` is analogous to
+[`ModelContextTool#description`](https://webmachinelearning.github.io/webmcp/#dom-modelcontexttool-description).
+
+The `toolautosubmit` [boolean attribute](https://html.spec.whatwg.org/C#boolean-attribute), lets the
```
bvandersloot-mozilla, line 49: I like this.

Here is where integration with visibility of form elements could also be interesting. If you require the user to review a form before clicking submit, why do you let agent fill things in that the user could not see.

Being in the control of the site, it can't be a prompt injection protection, but we can help a site build tools that an agent is less likely to misuse, and this fits that well.

declarative-api-explainer.md
```diff
@@ -0,0 +1,162 @@
...
+### Changes to form reset
+
+When a form is [reset](https://html.spec.whatwg.org/C#concept-form-reset) **OR** its tool
+declaration changes (as a result of `toolname` attribute modifications, for example), then any
+in-flight invocation of the tool will be cancelled, and the agent will be notified of this
+cancellation.
+
+### Input schema synthesis
+
+TODO: The exact algorithms reducing a form, its form-associated elements, and *their* attributes
```
bvandersloot-mozilla, line 108: Devil is in the details here, but careful specification/testing should be sufficient to get something right.

bvandersloot-mozilla (approved):
r+

declarative-api-explainer.md
```diff
@@ -0,0 +1,162 @@
...
+agent submit the form on the user's behalf after filling it out, without requiring the user to check
+it manually before submitting. If this attribute is missing when the agent finishes filling out the
+form, the browser brings the submit attribute into focus, and the agent should then tell the user to
+check the form contents, and submit it manually.
+
+When forms with these attributes are inserted, removed, or these attributes are updated, the form
+creates a new declarative WebMCP tool whose input schema is generated according to
+[Input schema synthesis](#input-schema-synthesis).
+
+We also introduce the new `toolparamname` and `toolparamdescription` attributes, which apply to form
```
beaufortfrancois: Chromium currently uses `toolparamtitle`, not `toolparamname`

declarative-api-explainer.md
```diff
@@ -0,0 +1,162 @@
...
+    }
+  ]
+});
+```
+
+... is equivalent to the following declarative form:
+
+```html
+<form toolname="search-cars" tooldescription="Perform a car make/model search" [...]>
+ <input type=text toolparamname="make" toolparamdescription="The vehicle's make (i.e., BMW, Ford)" required>
```
beaufortfrancois: FYI Here's what Chromium currently have:

- The optional `toolparamtitle` attribute maps to the json-schema property key. If omitted, the browser defaults to the input element’s name.
- The optional `toolparamdescription` attribute maps to the property description within the json-schema. In its absence, the browser uses the textContent of the associated <label> HTML element but skips descendants that are labelable or, if no label exists, the aria-description.

beaufortfrancois: I think you'd want to remove `toolparamname` for now, and use `name="model"` and `name="make"` instead.

declarative-api-explainer.md
```diff
@@ -0,0 +1,162 @@
...
+
+The `toolname` attribute is analogous to the imperative API's
+[`ModelContextTool#name`](https://webmachinelearning.github.io/webmcp/#dom-modelcontexttool-name),
+while `tooldescription` is analogous to
+[`ModelContextTool#description`](https://webmachinelearning.github.io/webmcp/#dom-modelcontexttool-description).
+
+The `toolautosubmit` [boolean attribute](https://html.spec.whatwg.org/C#boolean-attribute), lets the
+agent submit the form on the user's behalf after filling it out, without requiring the user to check
+it manually before submitting. If this attribute is missing when the agent finishes filling out the
+form, the browser brings the submit attribute into focus, and the agent should then tell the user to
```
bwalderman: Was this meant to be submit button?

```suggestion
form, the browser brings the submit button into focus, and the agent should then tell the user to
```

declarative-api-explainer.md
```diff
@@ -0,0 +1,162 @@
...
+};
+```
+
+**`toolactivated` and `toolcanceled` events
+
+We introduce these new events that get fired against he `Window` object when a WebMCP tool is run,
+and when its invocation is canceled. Some open questions:
+
+> [!WARNING]
+> Should these events fire for imperative tool call invocations as well?
```
bwalderman: Are these being added so that web developers can do javascript things when a declarative tool is called? What would be the reason to do this vs. use the submit event handler with preventDefault()?

declarative-api-explainer.md
```diff
@@ -0,0 +1,162 @@
...
+
+> [!WARNING]
+> For declarative, should they be fired at `Window` or at the `<form>` that registered the tool in
```
bwalderman, line 198: My vote would be for firing on the element that registered the tool. This makes it a lot easier to associate event handlers with tools on pages that have many declarative tools.

declarative-api-explainer.md
domfarolino: Yeah, the idea is we'd use the target site's existing `<script type=application/ld+json>` as the tool's response  for the model. The developer wouldn't drop anything declarative-WebMCP-specific in there.

I think handling the case where no `<script application/ld+json>` is present is TBD, so maybe I'll tighten this section up and mention that.

declarative-api-explainer.md
domfarolino: Yeah, I made this slightly stronger, but still said it was TBD. Hope you're happy with this!

declarative-api-explainer.md
domfarolino: Yep, I think I agree with this! Mind if we leave it as a WARNING for now since it reflect's Chromium's behavior, and then we can get rid of the warning once Chromium starts firing them at the element? I just want to make sure things don't get too out of sync.

declarative-api-explainer.md
andruud: +1

https://github.com/webmachinelearning/webmcp/pull/125

andruud: > FYI Here's what Chromium currently have:
> 
> * The optional `toolparamtitle` attribute maps to the json-schema property key. If omitted, the browser defaults to the input element’s name.

We don't fall back to the name, we just omit the `title` field in that case.

I think we probably don't need the `toolparamtitle` attribute at all. It's not currently part of the proposal, and I suggest keeping it that way. (We will still want to synthesize the `title` field in some cases, but we can do that based on `<label>`, etc.)

declarative-api-explainer.md
domfarolino: One complicating factor: if we consider firing these events for the imperative API as well, it becomes confusing that they sometimes fire on `Window`, and sometimes fire on `<form>` elements. I think it we had a dedicated `Tool` interface that inherited from `EventTarget` to represent imperative tools, then we could have a policy "these events fire directly on the tool: `Tool` for imperative tools, and `<form>` for declarative tools". Without such an interface, it might be weird to separate the firing. So let's discuss in a separate issue.

domfarolino: Filed https://github.com/webmachinelearning/webmcp/issues/126. Will resolve this for now.

declarative-api-explainer.md
domfarolino: ~Hmm, I don't actually see how inferring the output schema from a declarative form tool is possible. A form with complex inputs "search for shoes" gives the browser really no hint as to what structure the schema resulting from the search might take on the subsequent page. We'll probably need a new attribute for this? Though it feels odd to stuff a bunch of JSON into an HTML attribute 🤮~

Oops, I misread "impossible" as "possible". Yes I totally agree with everything here. Let's move this discussion to the linked-to issue #9 and continue over there!

declarative-api-explainer.md
domfarolino: Ah. I didn't see #125 (a PR on top of a PR!) before addressing this review feedback here, locally. Hmm, I've incorporated the link to https://json-schema.org/draft/2020-12/json-schema-validation#name-title-and-description from that PR into this one, and to streamline things I'll go ahead and close #125 and give @andruud co-author commit credit for this PR when it lands, if that's OK. Resolving this for now.

domfarolino: >  It's not currently part of the proposal, and I suggest keeping it that way.

Does this mean Chromium is getting rid of the `toolparamtitle` attribute? Otherwise, we certainly don't want Chrome to ship it, but the spec exclude it, right?

declarative-api-explainer.md
domfarolino: Agreed, we'll have to discuss this in more detail, since there are lots of attributes that can contribute to the input schema in various ways. As we discussed over a meeting, exactly how much should be specified is up for grabs (I personally lean towards spec'ing more of it). But yeah, we'll discuss more soon!

declarative-api-explainer.md
domfarolino: +1 ! Resolving, but lmk if you want to discuss anything further here.

declarative-api-explainer.md
domfarolino: Right, this lets the developer do things in response to a declarative form being executed, i.e., filled out with an agent's output, but **before submission**. So when the `toolautosubmit` attribute is not present, it gives the developer a hook to do some action to bring the form into view or otherwise alert the user that the form is ready for them to fill out, before submission.

We're also introducing CSS pseudo-classes that apply to a form that let styles react to agent-filled-out forms, and this is basically that JS version of that. I added the CSS pseudo-classes to the explainer in https://github.com/webmachinelearning/webmcp/pull/76/changes/53388c87ba372de6be84d5eb30a436c07d41944b.

declarative-api-explainer.md
```diff
@@ -0,0 +1,202 @@
...
+
+Chromium is implementing a loose version of this and will conduct testing/trials to see if what
+we've come up with should be supported by the community as a general approach.
+
+### Getting the form response to the agent
+
+When a form element performs a navigation, the first `<script type=application/ld+json>` tag on the
+target page is used as the cross-document tool's "response" that gets sent to the model.
+
+When no such a tag is present, probably we'll decide that the page's entire contents is sent to the
```
MiguelsPizza, line 127: Any thoughts on defaulting to a simple "form submitted successfully" message as the tool response? The signal to noise ratio of raw HTML can be huge and I'd argue it's better not to send it at all rather than send all of it. 

"Accurate semantic representation" and "useful to the model" aren't the same thing.  A content-heavy result page could easily blow out context in a way that's hard to predict or debug.

declarative-api-explainer.md
```diff
@@ -0,0 +1,202 @@
...
+
+1. New attributes that augment `<form>`s and [form-associated
+   elements](https://html.spec.whatwg.org/#form-associated-element), that expose these as WebMCP
+   tools to agents.
+2. Algorithms that deterministically "compile" a form and its associated inputs down to a WebMCP
+   "input schema", so that the agent knows how to fill out the form and submit it.
+3. Two ways of getting a form response back to the agent that invoked the form tool:
+    1. `SubmitEvent#respondWith()`, which lets JavaScript on the page override the default form
+       action, and pipe a response back to the agent without navigating the page.
+    2. Extracting `<script type="application/json-ld">` tags on the page that the form navigated to,
```
MiguelsPizza, line 32: I had some time to try out the `JSON-LD` cross document ergonomics this week and I'm leaning towards declarative tools not concerning themselves with cross-document outputs at all. I'd be curious to hear others' thoughts on this. The cross-document concern feels a bit self-inflicted and I found myself bypassing it in favor of read tools on the navigated-to page that the model can call on follow ups.

The model just needs to know: (a) did I fill the form correctly, if not what are the validation errors? (this does not cause a navigation) and (b) did it submit successfully 
(this could just be a generic message). From there it can follow up with read tools on the destination page to verify context.

This would let us drop concerns around output schema for declarative cross document tools as well.

+1 to @bwalderman's `<context>` idea. That or something like it feels like the right primitive here rather than a cross-document output API bolted onto forms. This sort of thing would be useful in general rather than as a specific solution for form submissions 
that cause navigations.

bwalderman: I also have concerns about `<script type="application/json-ld">`. The `<script>` tag isn't exactly intended for this purpose to begin with, and it's being used only to solve for cross-document form submission. That's just one specific case of a more general problem: What happens when an agent needs to initialize or catch-up state, either because it started browsing a new page, or it used a tool (imperative or declarative) that navigated. Read tools are important for the agent to get this initial state after navigation, so I'm leaning towards figuring out the declarative approach for read tools, and then letting this be the solution for cross-document form submission. This is the reason why I brought the `<context>` tag from the VOIX framework to everyone's attention. We'll want something like it anyway for the declarative approach to be considered complete, and it also happens to be a reasonable solution for cross-document form submission.

It may not end up being called `<context>`. It would probably be `<resource>` to keep aligned with MCP, but the semantics would be the same.

declarative-api-explainer.md
andruud: @domfarolino Indeed, we wouldn't ship that attribute if it doesn't end up in the spec. The prototype in Chromium supports it right now only because it was part of the original/early proposal.

domfarolino: Right, but since you're saying we shouldn't include it here, I'm wondering if you're recommending that because Chrome plans to rip it out of our implementation? Or do you think Chrome will keep it, and we should eventually put it in our spec proposal?

declarative-api-explainer.md
andruud: I also found the idea of vacuuming up "circumstantial" `JSON-LD` post-navigation and declaring it a "result" to a specific tool execution very questionable.

>  I'm leaning towards figuring out the declarative approach for read tools, and then letting this be the solution for cross-document form submission.

@bwalderman In that world, the _result_ of a declarative tool execution would indeed be something like "form submitted successfully" like @MiguelsPizza says? And then other side-effects of the tool execution would be understood by the agent from "read tools" it would anyway execute after any navigation?

declarative-api-explainer.md
andruud: I would like to rip it out, because there is something to be said for keeping this API as "low footprint" as possible. I understand that the author needs an "escape hatch" when `<label>` / ARIA is still insufficient to lead the agent on the right path (`toolparamdescription`). However, I don't think I see a need for _two_ such escape hatches? HTML forms should already be reasonably semantically equipped to provide a meaningful title for things via `<label>` (etc).

```
<input ... toolparamtitle="foo" toolparamdescription="bar">
```

would synthesize to two free-text, author-provided fields right next to each other:

```
{
    // ...
    "title":"foo",
    "description":"bar"
}
```

If `"description":"bar"` alone is not enough to guide the agent, then it seems to me like the author might just as well send `"description":"Description that covers both foo and bar"`, instead of having to worry about what passes as a "title" and what passes as a "description".

> Chrome plans

This is just my opinion right now, so I'm not sure if that amounts to a "Chrome plan" just yet. We would presumably follow whatever the group decision is. Until then we should be free to suggest a "default" in our proposal, and I'm suggesting that we omit it until someone asks for it. :slightly_smiling_face:

declarative-api-explainer.md
```diff
@@ -0,0 +1,202 @@
...
+```js
+[Exposed=Window]
+interface SubmitEvent : Event {
+  // ...
+  readonly attribute boolean agentInvoked;
+  undefined respondWith(Promise<any> agentResponse);
+};
+```
+
+**`toolactivated` and `toolcanceled` events
```
khushalsagar, line 175: when is the toolcanceled event dispatched?

declarative-api-explainer.md
bwalderman: @andruud that is my interpretation as well. Rather than try to work around loss of state due to navigation, accept this as a limitation and let a simple "form success" message be a cue to the agent that it should pull new state.

declarative-api-explainer.md
```diff
@@ -0,0 +1,202 @@
...
+
+With WebMCP, agents can perform complex actions like booking a flight or reserving a table by
+hooking into a site's own code designed to perform those actions, instead of the agent having to
+figure it out manually through a brittle series of screen shots, scrolls, and out-of-date screen
+reads.
+
+However, not all site functionality is exposed via JavaScript functions, and features that *are*
+take some effort to rewrite with an agent invoker in mind. Much of a site's functionality is
+provided via semantic HTML elements like `<form>`, and its various inputs. To **make it easier** for
+developers to expose this kind of site functionality while still using thte semantic web, we
```
yoavweiss: s/thte/the/

declarative-api-explainer.md
domfarolino: OK I think this sounds good to me. I'll leave this section alone then: we'll have `name` contributing the name of each property, `toolparamdescription` contributing the description, with fallbacks if the attribute is omitted (those fallback details will be filled in later).

So far, nothing will contribute to the `title` in the synthesized JSON schema, and I'll work towards removing `title` processing in Chrome.

declarative-api-explainer.md
domfarolino: See https://github.com/webmachinelearning/webmcp/pull/76#discussion_r2905853437 and that whole thread, on why I think this can be closed now.

declarative-api-explainer.md
domfarolino: OK I've filed https://github.com/webmachinelearning/webmcp/issues/135 and made a note in the explainer to follow-up over there. I think we're good to resolve this now.

declarative-api-explainer.md
domfarolino: Yep, this makes sense to me. I filed https://github.com/webmachinelearning/webmcp/issues/135 so we can has it out over there. See https://github.com/webmachinelearning/webmcp/pull/76#discussion_r2906201027. I'll resolve this now.

declarative-api-explainer.md
domfarolino: I added a note saying that we fire it only when the agent cancels the tool execution, and not when the developer implicitly does so by changing the form structure, or removing the form. I think that matches [Chromium's behavior](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/script_tools/model_context.cc;l=340-341;drc=18ae39a0c878c36745294b9980f9f4fc6c6b47f8), and is good enough for now.

I am actually curious about how an agent would cancel an active form tool though. I think Chromium fills the tool out synchronously in one shot, so I'm not sure where the opportunity for the agent to interrupt the "running" of the form tool would even be. @khushalsagar or @mfreed7, do you know?

domfarolino:
Alright, I've filed a few follow-up issues to discuss the remaining items in this PR elsewhere, and I think this PR is good to go. I plan on merging it sometime today, and following up on the other issues for remaining specific points.

declarative-api-explainer.md
andruud: > So far, nothing will contribute to the title in the synthesized JSON schema

I would probably argue that the JSON Schema `title` should be populated with `<label>` (etc) from regular HTML, and the JSON Schema `description` should be populated with `toolparamdescription`, falling back to ARIA.

Example:

```html
<label>
  Amount
  <input type=number name=amount toolparamdescription="In EUR, recurring monthly">
</label>
```

Becomes

```json
{
  "title":"Amount",
  "description":"In EUR, recurring monthly",
  "type":"number"
}
```

Today, _`description`_ falls back to `<label>`, which means we lose the `<label>` information if a `toolparamdescription` is provided.

andruud: Though that is a separate discussion from "should we keep the `toolparamtitle` attribute".

NgoiSigma:
Hi François and the Chrome WebMCP Team,

Thank you for the prompt update and the updated EPP documentation.

We are currently integrating WebMCP into the web interface of our local AI orchestrator (the NOVEYA project) and will adjust our frontend code to remove the toolparamtitle attribute accordingly.

We fully support the WebML CG's direction to potentially rely on the standard HTML <label> element in the future. This approach aligns perfectly with standard DOM architecture and will make integration much smoother for React-based environments like ours.

Best regards

PR #77, opened 2026-02-05
Adds a Makefile to improve Bikeshed quality of life.

markafoltz:
Copied from another BS repo.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/77.html" title="Last updated on Feb 10, 2026, 4:54 PM UTC (cd5554b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/77/dd5f397...cd5554b.html" title="Last updated on Feb 10, 2026, 4:54 PM UTC (cd5554b)">Diff</a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ ifdef LOCAL_BIKESHED
 .PHONY: lint watch
 
 lint: index.bs
-	bikeshed --print=plain --dry-run --force spec --line-numbers $<
+	bikeshed --print=plain --dry-run --die-when=late --line-numbers spec $<
 
 watch: index.bs
 	@echo 'Browse to file://${PWD}/index.html'
--- a/index.bs
+++ b/index.bs
@@ -19,6 +19,66 @@ Indent: 2
 Die On: warning
 </pre>
 
+<style>
+/* domintro and XXX from https://resources.whatwg.org/standard.css */
+.domintro {
+  position: relative;
+  color: green;
+  background: #DDFFDD;
+  margin: 2.5em 0 2em 0;
+  padding: 1.5em 1em 0.5em 2em;
+}
+
+.domintro dt, .domintro dt * {
+  color: black;
+  font-size: inherit;
+}
+.domintro dd {
+  margin: 0.5em 0 1em 2em; padding: 0;
+}
+.domintro dd p {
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,23 @@
...
+	curl https://api.csswg.org/bikeshed/ -f -F file=@$< >$@;
+else
+	bikeshed spec
+endif
+
+ifdef LOCAL_BIKESHED
+.PHONY: lint watch
+
+lint: index.bs
+	bikeshed --print=plain --dry-run --force spec --line-numbers $<
```
domfarolino: `--force` is die on _nothing_, right? We probably don't want that. What does spec-prod use by default @tabatkins?

tabatkins: I don't actually know. I *assume* it always succeeds, aka `--force`. But for a lint command I definitely think you should use a more restrictive value, that's kinda the point of linting.

Btw `--line-numbers` no longer does anything, I need to remove that flag. Bikeshed always generates good line numbers now.

beaufortfrancois: FWIW WebGPU uses `DIE_ON=everything` in https://github.com/gpuweb/gpuweb/blob/main/spec/Makefile#L12 and https://github.com/gpuweb/gpuweb/blob/main/tools/invoke-bikeshed.sh#L28

anssiko: For consistent behaviour between local builds and https://github.com/webmachinelearning/webmcp/blob/main/.github/workflows/auto-publish.yml#L17 we should die on warnings i.e. everything.

markafoltz: For the `lint` target we never want to abort processing early and output as many errors/warning as possible in a single pass.

Outputting the spec is optional as `lint` is a phony target that will always run.

I see that bikeshed now supports `--die-when=late`.  Does that continue processing (even after fatal errors)?

markafoltz: After some local testing (and reading of the bikeshed source) this combination of arguments should do it.

tabatkins: Bikeshed now *defaults* to `Die When: late`, so all the errors will be output before it eventually fails to actually generate the output file.

Just checking - was there something lacking in the docs that caused you to go read the source?

Issue #92, opened 2026-02-17
Who owns the validation layer?


MiguelsPizza:
Currently, `inputSchema` is purely a semantic hint for the agent. If the agent calls the tool incorrectly, it is up to the tool creators to validate and provide feedback to the caller. From my experience, this feedback is really important for SLM's (prompt API) that often miss-call tools on the first attempt and rely on the validation error feedback to self correct.

Considering that this is something every tool creator is going to have to maintain individually, it would be nice if the browser instead did the validation and returned validation errors in a structured way ([prior](https://www.jsonrpc.org/specification#error_object) [art](https://standardschema.dev/schema))

Related security work that could benefit from browser owned validation:  [#45](https://github.com/webmachinelearning/webmcp/issues/45), [#73](https://github.com/webmachinelearning/webmcp/issues/73) and [#9](https://github.com/webmachinelearning/webmcp/issues/9)

bwalderman:
Validating the input schema at the browser level makes sense as a default behavior. In the case of declarative `<form>` tools, it might even make sense to map input schema errors to HTML form validation errors that the user would then have a chance to correct.

I can see developers wanting to customize error behavior and the error message that ultimately gets sent back to the agent. Surfacing errors as an `error` event on the `modelContext` object with details about the tool call and the specific validation errors would give the developer a chance to intercept the error and do things like: log it, rethrow a more descriptive error, or maybe even attempt recovery and let the tool call proceed. If an `error` event isn't handled, the browser can just do the default behavior which would be to return the raw schema validation results back to the agent.

johannhof:
I agree that this makes sense to do on browser side

domfarolino:
As far as I can tell, there are **three** types of "validation" we can talk about:

1. **Meta validation**: of the `inputSchema`. This is basically: does `inputSchema` use the correct JSON schema structure and vocabulary, like `anyOf`, `properties`, `min`, etc?
2. **Input validation**: of the model's output against the `inputSchema` that the execute callback expects
3. **Output validation**: that the execute callback's return value matches the pre-declared output schema, when https://github.com/webmachinelearning/webmcp/issues/9 is resolved.

We should probably do all of the above, but some notes:

Meta validation doesn't seem too bad, but we have to make a lot of decisions. For example, from what I can tell `{}` is a valid JSON Schema, but that's not very useful. So:

[...]

Issue #51, opened 2025-11-03
Define the API for in-page Agents to use a site's declared tools


khushalsagar:
The [modelContext](https://github.com/webmachinelearning/webmcp/blob/main/docs/proposal.md#modelcontext) API currently only provides APIs for a site to declare tools. We also need an API surface for an Agent embedded on the site to use these tools. Technically this could all be managed within the site but having a dedicated web API is better for multiple reasons. https://github.com/webmachinelearning/webmcp/issues/43#issuecomment-3478492067 has a detailed explanation, adding a quick summary below:

- We can a common registry of tools from all sources for all Agents.
- The browser can help mediate which Agent is accessing the site's functionality at a time.

We should first resolve on the fact that this API is necessary.

In terms of API details, a few things on my mind:

- It would be nice to have one interface for declaring functionality (right now tools, likely context going forward) since that's shared across all Agents. And a separate one for an Agent to use this functionality.
- Should we assume that there's only one "in-page Agent" on a site? That's related to the agentIdentifier question on https://github.com/webmachinelearning/webmcp/issues/44. The browser can either allocate a "in-site agent" identifier if the site can only have one agent. Or we need an API for the site to request these.
- We also added an interface called `agent` for the dependencies the site needs. I now think these are more like user-agent/browser dependencies. The site asks the browser for something which *may* need to ask the Agent.

43081j:
isn't this just _the_ web API we're proposing overall?

i.e. `navigator.modelContext` will have a means to call one of the registered tools and all agents, in page or not, will call that

maybe the name of this issue doesn't need to include "for in-page Agents" since its just the tool calling API overall, for all agents.

khushalsagar:
Meeting notes from TPAC discussion [here](https://www.w3.org/2025/11/11-webmachinelearning-minutes.html#5dde).

> isn't this just the web API we're proposing overall?

No, the [API](https://github.com/webmachinelearning/webmcp/blob/main/docs/proposal.md#api) resolved so far is sufficient for Agents accessing these tools through browser internal APIs (browser's built-in Agent, extensions etc). That's a detail specific to each browser implementation, it doesn't need to be defined by the spec.

43081j:
[...]

PR #113, opened 2026-02-26
Add basic tool infra and complete `registerTool()`

domfarolino:
This PR adds some basic tool infrastructure and fills out the `registerTool()` method properly. After this PR, I will upload follow-ups for `unregisterTool()` and more infrastructure to expose/verbalize tools definitions to a model (much will be implementation-defined, since we're not defining the actual protocol through which this is carried out), and handle tool execution.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -107,13 +107,13 @@ A <dfn>model context</dfn> is a [=struct=] with the following [=struct/items=]:
 
 <dl dfn-for="model context">
   : <dfn>tool map</dfn>
-  :: a [=map=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=tool data=]
+  :: a [=map=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=tool definition=]
      [=structs=].
 </dl>
 
-A <dfn>tool data</dfn> is a [=struct=] with the following [=struct/items=]:
+A <dfn>tool definition</dfn> is a [=struct=] with the following [=struct/items=]:
 
-<dl dfn-for="tool data">
+<dl dfn-for="tool definition">
   : <dfn>name</dfn>
   :: a [=string=] uniquely identifying a tool registered within a [=model context=]'s [=model
      context/tool map=]; it is the same as the [=map/key=] identifying this object.
@@ -266,24 +266,24 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>)</dfn> method step
 1. Let |read-only hint| be true if |tool|'s {{ModelContextTool/annotations}} [=map/exists=], and if
    its {{ToolAnnotations/readOnlyHint}} [=map/exists=] and is true. Otherwise, let it be false.
 
-1. Let |tool data| be a new [=tool data=], with the following [=struct/items=]:
+1. Let |tool definition| be a new [=tool definition=], with the following [=struct/items=]:
 
-   : [=tool data/name=]
+   : [=tool definition/name=]
    :: |tool name|
 
-   : [=tool data/description=]
+   : [=tool definition/description=]
    :: |tool|'s {{ModelContextTool/description}}
 
-   : [=tool data/input schema=]
+   : [=tool definition/input schema=]
    :: |stringified input schema|
 
-   : [=tool data/execute steps=]
+   : [=tool definition/execute steps=]
    :: steps that invoke |tool|'s {{ModelContextTool/execute}}
 
-   : [=tool data/read-only hint=]
+   : [=tool definition/read-only hint=]
    :: |read-only hint|
 
-1. Set [=this=]'s [=ModelContext/internal context=][|tool name|] to |tool data|.
+1. Set [=this=]'s [=ModelContext/internal context=][|tool name|] to |tool definition|.
 
 </div>
 


index.bs
```diff
@@ -181,9 +226,64 @@ The <dfn method for=ModelContext>clearContext()</dfn> method steps are:
...
 
-1. TODO: fill this out.
+1. Let |tool map| be [=this=]'s [=ModelContext/internal context=]'s [=model context/tool map=].
+
+1. Let |tool name| be |tool|'s {{ModelContextTool/name}}.
+
+1. If |tool map|[|tool name|] [=map/exists=], then [=exception/throw=] an {{InvalidStateError}}
+   {{DOMException}}.
+
+1. If either |tool name| or {{ModelContextTool/description}} is the empty string, then
```
beaufortfrancois, line 238: Shall we also raise an error if either are `null`?

domfarolino: No need since they are marked as `required` by the Web IDL, so they can't make it this far and be null.

beaufortfrancois: @domfarolino Is there a bug then in Chrome? This code doesn't raise any error:
```js
navigator.modelContext.registerTool({name: null, description: null, execute: () => {}})
```

beaufortfrancois: And somehow calling twice raises:
```
Uncaught InvalidStateError: Failed to execute 'registerTool' on 'ModelContext': Duplicate tool name
    at <anonymous>:1:24
```

domfarolino: When the JavaScript value `null` converts to a DOMString via `String(null)`, you get the literal string `"null"`, so actually this is WAI, even though it's weird. So when you register a tool with the name `null`, it's registering under the name `"null"`, which is why registering it again finds a duplicate tool

This behavior is elsewhere on the platform too. Try `el.addEventListener(null, () => console.log('called'))` and look at the event handler that gets added in the DevTools event handler pane. You'll find that a handler for the `null` event is added 🤯.

index.bs
```diff
@@ -95,14 +101,49 @@ A <dfn>browser’s agent</dfn> is an [=agent=] provided by or through the browse
...
+A <dfn>model context</dfn> is a [=struct=] with the following [=struct/items=]:
 
-<h2 id="accessibility">Accessibility considerations</h2>
+<dl dfn-for="model context">
+  : <dfn>tool map</dfn>
+  :: a [=map=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=tool data=]
+     [=structs=].
+</dl>
+
+A <dfn>tool data</dfn> is a [=struct=] with the following [=struct/items=]:
```
bwalderman: I'd suggest "tool definition" instead. Even though "data" is often used as if it's singular, it's plural so this reads a little awkward to me.

index.bs
```diff
@@ -95,14 +101,49 @@ A <dfn>browser’s agent</dfn> is an [=agent=] provided by or through the browse
...
+
+<dl dfn-for="tool data">
+  : <dfn>name</dfn>
+  :: a [=string=] uniquely identifying a tool registered within a [=model context=]'s [=model
+     context/tool map=]; it is the same as the [=map/key=] identifying this object.
+
+  : <dfn>description</dfn>
+  :: a [=string=].
+
+  : <dfn>input schema</dfn>
```
bwalderman, line 124: Is there a particular reason this is being stored as a string in the tool map, and not an object? It seems like this is an optimization for passing the schema to an LLM, but here in the spec we'll likely have algorithms that need to retrieve the original object and read from it (e.g. input validation, listTools).

index.bs
```diff
@@ -143,6 +184,10 @@ interface ModelContext {
 };
 </xmp>
 
+Each {{ModelContext}} object has an associated <dfn for=ModelContext>internal context</dfn>, which
```
bwalderman, line 187: What is the internal context for?

bwalderman (changes requested):

index.bs
beaufortfrancois: Thanks for the explanation!

index.bs
domfarolino: The `ModelContext` IDL definition is just a web-exposed JavaScript interface, and so to store internal state associated with it (like the tool map from string => tool dfn), we need some internal data structure attached to it. Right now the only thing we really need to store associated with a given `ModelContext` interface is a tool map, but we'll need more state as well like things that track the current tool execution/pending executions, and possibly more. So I'm creating a general "internal context" bag for all of this stuff for now, and stuffing the tool map in it. This is a pretty common pattern for web specs in general.

index.bs
domfarolino: This is a good point. On one hand, it feels too early to stringify the input schema immediately upon registration for storage here, especially if we want to support an API that enumerates tools later. This is in fact [a limitation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/script_tools/model_context_testing.idl;l=8;drc=df2f2ac216704ff5f2d3e7ff77fba0c9cffbc133) of the `ModelContextTesting` interface (test-only) in Chromium (which annoyingly exposes a registered tool's input schema as a `DOMString` instead of an `object`).

It's tempting to store the Web IDL `object` here and only serialize it when the model wants it. However, the timing of "tool registration" and "model consumption" is unclear—at the moment, Chromium operates on a "pull" model, but is considering moving to a "push" model, where tools are sent to the model upon registration, not on each conversation turn. All of this is to say, we probably want to perform tool serialization—to make the tool schema and everything ready for model consumption—right away at registration time since serialization can fail, and we want to give developers that feedback immediately.

I can't think of a better way to do this besides storing the Web IDL `object` type here, _alongside_ the stringified serialization of the very same object. The `object` representation would be used for any validation against model output (see #92), and for any future "list tool" API (see #51). The stringified serialization would be used for model consumption, when I get around to spec'ing that next.

So since we don't have validation yet (see #92), and we don't have a "list tools" API yet (see #51), my slight preference would be to keep this PR as-is, and in a follow-up PR add a sibling member that stores the actual raw `object` alongside it as the need arises.

How does that sound?

bvandersloot-mozilla: > All of this is to say, we probably want to perform tool serialization—to make the tool schema and everything ready for model consumption—right away at registration time since serialization can fail, and we want to give developers that feedback immediately.

My prototyping in Firefox ran in to and matched this reasoning independently. 

> How does that sound?

+1 here.

bwalderman: Sounds good.

PR #124, opened 2026-03-03
Remove Chromium note in serialization algorithm

beaufortfrancois:
Following https://chromium-review.googlesource.com/c/chromium/src/+/7626740, we should remove the Chromium note in the serialization algorithm.


beaufortfrancois:
@domfarolino Please review

PR #125, opened 2026-03-04
[explainer] Remove toolparamname attribute

andruud:
We should likely use the standard `name` attribute here, and just drop the `toolparamname` attribute. At the very least, the input elements in the example should *also* have a `name` attribute to be more realistic.

Also fix some typos/mistakes.


andruud:
cc @domfarolino; looks like I can't set assignees/reviewers here.

domfarolino:
Thanks! See https://github.com/webmachinelearning/webmcp/pull/76#discussion_r2884317127 for how this got incorporated into the base PR—will give git co-author credit over there, and close this PR to streamline things.

PR #146, opened 2026-03-19
Add ToolActivatedEvent and ToolCancelEvent interfaces

beaufortfrancois:
Following https://github.com/webmachinelearning/webmcp/issues/126#issuecomment-4001724830, this PR adds new `ToolActivatedEvent` and `ToolCancelEvent` interfaces.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/webmcp/pull/146.html" title="Last updated on Aug 6, 2026, 9:49 AM UTC (1099a80)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/146/0967151...beaufortfrancois:1099a80.html" title="Last updated on Aug 6, 2026, 9:49 AM UTC (1099a80)">Diff</a>

PR #164, opened 2026-04-16
Add basic page observation infrastructure

domfarolino:
This PR adds basic "page observation" and scheduling infrastructure, to lay the ground work for browser-agent-invoked tool execution. After this PR, I'll create one for tool execution, to unblock https://github.com/webmachinelearning/webmcp/pull/146.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/164.html" title="Last updated on Apr 23, 2026, 9:48 PM UTC (aded79c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/164/1380910...aded79c.html" title="Last updated on Apr 23, 2026, 9:48 PM UTC (aded79c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -447,8 +447,8 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
 
 A web site's functionality, as registered with the APIs in this specification, is exposed.
 
-The [=user agent=]'s [=browser's agent=] runs [=in parallel=] to any [=event loops=] associated
-with a {{ModelContext}} [=relevant global object=]. Steps running on the [=browser's agent=] get
+The [=user agent=]'s [=browser agent=] runs [=in parallel=] to any [=event loops=] associated
+with a {{ModelContext}} [=relevant global object=]. Steps running on the [=browser agent=] get
 queued on its <dfn>AI agent queue</dfn>, which is the result of [=starting a new parallel queue=].
 
 Conversely, steps queued *from* the [=browser agent=] onto the [=event loop=] of a given
--- a/index.bs
+++ b/index.bs
@@ -466,7 +466,7 @@ The [=browser agent=], on the other hand, does not run JavaScript on the page. I
 view of the page's tools and any other relevant context by getting an [=observation=]. An
 <dfn>observation</dfn> is an [=implementation-defined=] data structure containing at least a <dfn
 for=observation>tool map</dfn>, which is a [=map=] whose [=map/keys=] are [=Document/unique ID=]s,
-and whose [=map/values=] are [=tool definitions=].
+and whose [=map/values=] are [=lists=] of [=tool definition=] [=structs=].
 
 Note: An [=observation=] is usually a "snapshot" distillation of a page being presented to the user,
 along with any other state the [=user agent=] believes is relevant for the [=browser agent=]; this
--- a/index.bs
+++ b/index.bs
@@ -468,10 +468,10 @@ for=observation>tool map</dfn>, which is a [=map=] whose [=map/keys=] are [=Docu
 and whose [=map/values=] are [=tool definitions=].
 
 Note: An [=observation=] is usually a "snapshot" distillation of a page being presented to the user,
-along with any other state the [=user agent=] believes is relevant for the [=browser agent=]. See
-[Annotated Page Content
+along with any other state the [=user agent=] believes is relevant for the [=browser agent=]; this
+often includes screenshots of the page, not just a DOM serialization. See [Annotated Page Content
 (APC)](https://chromium.googlesource.com/chromium/src.git/+/main/third_party/blink/renderer/modules/content_extraction/readme.md)
-in the Chromium project.
+in the Chromium project for an example of what might contribute to an observation.
 
 <hr>
 
--- a/index.bs
+++ b/index.bs
@@ -458,6 +458,12 @@ global object=]'s <dfn noexport>tool calling task source</dfn>.
 
 <h3 id="observations">Page observations</h3>
 
+<em>This section is non-normative. It contains an example of infrastructure that a [=user agent=] might
+employ to expose a tab's tools to a [=browser agent=], and illustrates how that infrastructure
+interacts with the web platform.</em>
+
+<hr>
+
 In-page [=agents=] implemented in JavaScript can "observe" the tools that a page offers by using the
 {{ModelContext}} APIs directly, and any other platform APIs to obtain necessary context about the
 page in order to actuate it appropriately.
--- a/index.bs
+++ b/index.bs
@@ -498,6 +498,10 @@ steps:
         object=]'s {{Navigator}}'s [=Navigator/modelContext=]'s [=ModelContext/internal context=]'s
         [=model context/tool map=]'s [=map/values=], which are [=tool definitions=].
 
+1. Perform any [=implementation-defined=] steps to add anything to |observation| that the [=user
+   agent=] might deem useful or necessary, besides just populating the [=observation/tool map=].
+   This might include annotated screenshots of the page, parts of the accessibility tree, etc.
+
 1. Perform any [=implementation-defined=] steps with |observation| and the [=browser agent=], to
    expose the |observation|'s [=observation/tool map=] to the [=browser agent=] in whatever way it
    accepts.
--- a/index.bs
+++ b/index.bs
@@ -460,7 +460,7 @@ global object=]'s <dfn noexport>tool calling task source</dfn>.
 
 <em>This section is non-normative. It contains an example of infrastructure that a [=user agent=] might
 employ to expose a tab's tools to a [=browser agent=], and illustrates how that infrastructure
-interacts with the web platform.</em>
+interacts with the web platform, for the purposes of implementer guidance.</em>
 
 <hr>
 
--- a/index.bs
+++ b/index.bs
@@ -454,7 +454,7 @@ queued on its <dfn>AI agent queue</dfn>, which is the result of [=starting a new
 
 Conversely, steps queued *from* the [=browser agent=] onto the [=event loop=] of a given
 {{ModelContext}} object (i.e., the "main thread" where JavaScript runs) are queued on its [=relevant
-global object=]'s [=AI task source=].
+global object=]'s <dfn noexport>tool calling task source</dfn>.
 
 <h3 id="observations">Page observations</h3>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -439,6 +441,89 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
 }
 </pre>
 
+<h2 id="interaction-with-agents">Interaction with agents</h2>
+
+<h3 id="event-loop">Event loop integration</h3>
+
+A web site's functionality, as registered with the APIs in this specification, is exposed.
```
markafoltz:
```suggestion
A web site's functionality is exposed as registered with the APIs in this specification.
```

index.bs
```diff
@@ -439,6 +441,89 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
...
+
+The [=user agent=]'s [=browser's agent=] runs [=in parallel=] to any [=event loops=] associated
+with a {{ModelContext}} [=relevant global object=]. Steps running on the [=browser's agent=] get
+queued on its <dfn>AI agent queue</dfn>, which is the result of [=starting a new parallel queue=].
+
+Conversely, steps queued *from* the [=browser agent=] onto the [=event loop=] of a given
```
markafoltz, line 455: Use `browser agent` or `browser's agent` consistently

index.bs
```diff
@@ -439,6 +441,89 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
...
+{{ModelContext}} object (i.e., the "main thread" where JavaScript runs) are queued on its [=relevant
+global object=]'s [=AI task source=].
+
+<h3 id="observations">Page observations</h3>
+
+In-page [=agents=] implemented in JavaScript can "observe" the tools that a page offers by using the
```
markafoltz, line 467: Nit: Agents could use WASM in principle, but JS will be involved to interact with `ModelContext` at some point so...

index.bs
```diff
@@ -439,6 +441,89 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
...
+{{ModelContext}} APIs directly, and any other platform APIs to obtain necessary context about the
+page in order to actuate it appropriately.
+
+The [=browser agent=], on the other hand, does not run JavaScript on the page. Instead, it obtains a
+view of the page's tools and any other relevant context by getting an [=observation=]. An
+<dfn>observation</dfn> is an [=implementation-defined=] data structure containing at least a <dfn
+for=observation>tool map</dfn>, which is a [=map=] whose [=map/keys=] are [=Document/unique ID=]s,
+and whose [=map/values=] are [=tool definitions=].
```
markafoltz: I would think this needs to be doubly keyed by (document_id, tool_name) at a minimum?

index.bs
```diff
@@ -439,6 +441,89 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
...
+
+Note: An [=observation=] is usually a "snapshot" distillation of a page being presented to the user,
+along with any other state the [=user agent=] believes is relevant for the [=browser agent=]. See
```
markafoltz: You might want to mention screenshots can be included (not just DOM state)

index.bs
```diff
@@ -439,6 +441,89 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
...
+[Annotated Page Content
+(APC)](https://chromium.googlesource.com/chromium/src.git/+/main/third_party/blink/renderer/modules/content_extraction/readme.md)
+in the Chromium project.
+
+<hr>
+
+<div algorithm>
+To <dfn>perform an observation</dfn> given a [=top-level traversable=] |traversable|, run these
```
markafoltz, line 486: I think this can be guessed from context, but these steps are only for populating the tool map, not other parts of the observation that are up to the implemenation?

index.bs
```diff
@@ -439,6 +441,89 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
...
+     security information associated with [=tool definitions=], such as the originating [=origin=],
+     among other things, so that the backing model has an idea of the different parties at play, and
+     can most safely carry out the end user's intent.
+
+</div>
+
+Each {{Document}} object has a <dfn for=Document>unique ID</dfn>, which is a [=unique internal
+value=].
+
+The times at which a [=browser agent=] [=performs an observation=] are [=implementation-defined=].
```
markafoltz, line 531: We may need to be more proscriptive here; if tools are registered or unregistered, or frames enter or leave the frame tree, those should trigger a re-observation.

index.bs
```diff
@@ -439,6 +441,89 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
...
+
+     1. Let |id| be |document|'s [=Document/unique ID=].
+
+     1. Set |observation|'s [=observation/tool map=][|id|] = |document|'s [=relevant global
+        object=]'s {{Navigator}}'s [=Navigator/modelContext=]'s [=ModelContext/internal context=]'s
+        [=model context/tool map=]'s [=map/values=], which are [=tool definitions=].
+
+1. Perform any [=implementation-defined=] steps with |observation| and the [=browser agent=], to
+   expose the |observation|'s [=observation/tool map=] to the [=browser agent=] in whatever way it
+   accepts.
```
markafoltz, line 514: I expect this will also update the tool map exposed to other frames in listTools() once that is spec'ed out? Otherwise I am wondering how we will measure interop since the side effect is only exposed to the browser agent.

index.bs
```diff
@@ -439,6 +441,89 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
...
 
+<h2 id="interaction-with-agents">Interaction with agents</h2>
+
+<h3 id="event-loop">Event loop integration</h3>
+
+A web site's functionality, as registered with the APIs in this specification, is exposed.
+
+The [=user agent=]'s [=browser's agent=] runs [=in parallel=] to any [=event loops=] associated
+with a {{ModelContext}} [=relevant global object=]. Steps running on the [=browser's agent=] get
+queued on its <dfn>AI agent queue</dfn>, which is the result of [=starting a new parallel queue=].
```
markafoltz, line 453:
```suggestion
queued on its <dfn>browser agent queue</dfn>, which is the result of [=starting a new parallel queue=].
```

markafoltz (commented):
A few nit picks, it looks like the spec generator API threw a few errors to clean up.

index.bs
domfarolino: Actually performing an observation will not have any web-observable effects. When I spec the Promise-returning `getTools()`, we'll have to move to a central browser process location of all combined tools in the frame tree. Then, the page observation logic here will move from doing what APC does in Chromium now (gathering all tools from the tree) to just taking a copy of that browser-side map, and packaging it up here in some way for the observation.

So in short, this is not the map that `getTools()` will consult. This is just the map for observations. And its instantiation will be simplified once `getTools()` exists imminently.

index.bs
domfarolino: Since we're using the [AI task source](https://webmachinelearning.github.io/writing-assistance-apis/#ai-task-source) in the renderer/Document's event loop, I'd like to keep the browser side equivalent also prefixed with "AI", to mirror this. Is that OK?

index.bs
domfarolino: I agree. My plan is this: when `getTools()` exists, whenever we queue tasks on the AI agent queue to update the browser process map of all tools, I'll also add a step about optionally performing observations *then*. So I was thinking of holding off on that until that infra is laid. How does that sound?

index.bs
domfarolino: Ah, good point. I added another step just before the last one, instructing the browser to add anything else besides just tools that it finds useful. That makes use of the other implementation-defined data I mention farther up that is also attached to an `[=observation=]`.

index.bs
domfarolino: That'd only be necessary if within a single Document, you could have multiple tools with the same name. But that shouldn't be possible, so keying on document ID means all of the tools associated with that document will be unique. Across multiple Documents, you can have duplicate tool names, but that should be taken care of. Does that sound right or am I missing something?

index.bs
domfarolino: True, but yeah this is just meant to allude to author-provided agents vs. not, so I think it's close enough.

index.bs
domfarolino: Done, thanks.

index.bs
domfarolino: Yikes, this whole thing was previously just a sentence I never finished it seems. Thanks. I went with `A web site's functionality is exposed to [=agents=] as tools that live in a [=Document=]'s [=event
loop=], that get registered with the APIs in this specification.` to add more scheduling lingo.

index.bs
markafoltz: It's not clear from reading the Writing Assistance spec that the AI task source in that spec is related to the one you are defining here, as the AI task source is used to fire events and resolve promises in site script.   Also, while it's implied that AI agents will be the main driver of script tools, it's not a requirement.

Fine as-is though for this PR.

index.bs
markafoltz: The nuance here I missed is that the value in the map is an array of tool definitions and not a single definition.  It may be worth being more explicit here.

index.bs
markafoltz: Thanks; I would consider putting the new step in a non-normative note.

I'm not sure it's necessary to write in normative spec a step that (for now) is proprietary to each browser/agent as that's not going to be directly observable to the platform.  On the other hand, it's useful for developers to know that agents will consume page information beyond tool output.

index.bs
markafoltz: Exactly what I was thinking!

index.bs
markafoltz: SGTM

markafoltz (approved):

index.bs
domfarolino: You didn't miss anything, I'm just realizing that this structure is incorrect. Right now it's `Document ID => single tool definition struct` which is a mistake; it should be a `Document ID => array of tool definition structs`. I've fixed that, thanks!

index.bs
domfarolino: You have a good point. Especially if we're using this task source to schedule the invocation of tools across frames which might have nothing to do with the built-in agent, this should be named something else. OK I introduced a new task source called "tool calling task source" instead of delegating to the WA spec's.

index.bs
domfarolino: Yeah that's reasonable. I made this section non-normative, and more tuned for implementer guidance. Done!

PR #177, opened 2026-05-15
Scope `ModelContext` to `Document`, and handle detached documents

### LINKED ISSUE #173: Make tools Document-scoped instead of Window-scoped
It's important that tools are document-scoped. Most of the time Window/Navigator and Document are 1:1, however [navigations away from the initial about:blank Document share a Window object](https://html.spec.whatwg.org/#windows:window), which complicates things.

The case where this is observable is when you register tools with the initial about:blank Document, and then navigate away from it. Per the current spec, after the navigation, calls to `navigator.modelContext` return the same object that holds the same map of tools, which means tools are effectively shared across this navigation.

I recently caught this bug in Chromium when working on the iframe + permission policy implementation, and ended up changing Blink's implementation such that:

1. The backing `ModelContext` interface object is stored on Document instead of Navigator
2. The `navigator.modelContext` IDL getter doesn't just return the same object; instead it consults its Window's active Document, and returns _its_ `ModelContext` object. It returns null if the Window isn't available (detached document) or if the document isn't fully active (bf-cache).

I'll update the spec to agree with this.

----

## Follow-up: API placement

An oddity of the above is that the API is effectively scoped to the Document, yet it lives off of `navigator`. I personally never liked putting things on `navigator`, and would recommend that we do one of two things:
 - Move it to `Document`, giving us `document.modelContext`. This seems more intuitive
 - Move it to `Window`. This has the same scoping oddity as `Navigator` since both are 1:1, but it follows the guidance in https://github.com/w3ctag/design-principles/issues/426 that most editors seem to have expressed, which is hanging things off of the global instead of a weird namespace like `navigator`

domfarolino:
This PR does two things:
 - Scopes the `ModelContext` object to `Document` instead of `Navigator`, and explains why.
 - Makes the `modelContext` attribute return null when the document is detached (by checking when the Window's browsing context is null). This matches the behavior of the `contentWindow` and [`contentDocument` getters](https://html.spec.whatwg.org/#concept-bcc-content-document)—return nothing, even if those objects are allocated and kept alive by other references, in the detached document case.

Note that this PR does not handle the "document not fully active" case, which covers detached documents *and* documents in the bf-cache. Non-fully active documents probably shouldn't impact the `modelContext` getter—it seems weird for getters like that to return different values based on whether the document is in the bf-cache—but rather need special behavior in the `registerTool()` method, and the coming `getTools()` and `executeTool()` methods.

This PR mostly addresses https://github.com/webmachinelearning/webmcp/issues/173, with the exception of considering moving the API to `self`, from `navigator`. As much as I want to do this, I'm sympathetic to https://github.com/w3ctag/design-principles/issues/426#issuecomment-3824607888, so for now I'm leaning towards not moving it. So I'll say that this PR closes https://github.com/webmachinelearning/webmcp/issues/173.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/177.html" title="Last updated on May 18, 2026, 8:24 PM UTC (f11b3da)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/177/a99ffd2...f11b3da.html" title="Last updated on May 18, 2026, 8:24 PM UTC (f11b3da)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -86,6 +86,14 @@ spec:html; type:dfn;
   text:browsing context group set
   text:unique internal value
 </pre>
+<pre class="anchors">
+spec: html; type: dfn; urlPrefix: https://html.spec.whatwg.org/C
+  text: create and initialize a Document object; url: initialise-the-document-object
+  # Because `text` and `url` are the same (modulo slashes), we should be able to
+  # exclude `url`. But the colon breaks the URL in `text`, forcing us to include
+  # `url`.
+  text:is initial about:blank; url: is-initial-about:blank
+</pre>
 
 <h2 id="intro">Introduction</h2>
 
@@ -112,6 +120,7 @@ A <dfn>model context</dfn> is a [=struct=] with the following [=struct/items=]:
   :: a [=map=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=tool definition=]
      [=structs=].
 </dl>
+
 
 A <dfn>tool definition</dfn> is a [=struct=] with the following [=struct/items=]:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -183,19 +192,40 @@ https://dlaliberte.github.io/bikeshed-intro/#a-strategy-for-incremental-developm
 
 <h3 id="navigator-extension">Extensions to the {{Navigator}} Interface</h3>
 
-The {{Navigator}} interface is extended to provide access to the {{ModelContext}}.
+Each {{Document}} object has an <dfn for=Document>associated {{ModelContext}}</dfn>, which is a
+{{ModelContext}} object.
+
+Upon creation of the {{Document}} object, its [=Document/associated ModelContext|associated
+<code>ModelContext</code>=] must be set to a [=new=] {{ModelContext}} object created in the
```
domfarolino, line 200: Hey @tabatkins, I could not for the life of me figure out how to add `{{ModelContext}}` or `<code>ModelContext</code>` to the actual linking text, so I had to settle on this super verbose `[=foo|<code>foo</code>=]` which is really redundant and I can't figure out if there's a way around it, so I can tighten these up. It's not blocking, so if there's no solution, that's fine.

tabatkins: There's no "proper" way around it - linking text is *text*, not markup.

There is a workaround, tho - use an internal macro:

```bikeshed
<pre class=metadata>
...
Text Macro: ASSOCIATEDMODELCONTENT [=Document/associated ModelContext|associated <code>ModelContext</code>=]
</pre>

...
Upon creation of the {{Document}} object, its [ASSOCIATEDMODELCONTENT] must be set to a [=new=] {{ModelContext}} object
```

(You could also scope it to just use the macro for the visible text part, but if you're doing this link a lot you might as well macro the whole thing.)

index.bs
```diff
@@ -183,19 +192,40 @@ https://dlaliberte.github.io/bikeshed-intro/#a-strategy-for-incremental-developm
...
 </xmp>
 
-Each {{Navigator}} object has an associated <dfn for=Navigator>modelContext</dfn>, which is a
-{{ModelContext}} instance created alongside the {{Navigator}}.
-
 <div algorithm>
-The <dfn attribute for=Navigator>modelContext</dfn> getter steps are to return [=this=]'s [=Navigator/modelContext=].
+The <dfn attribute for=Navigator>modelContext</dfn> getter steps are:
+
+1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is null, then return null.
```
beaufortfrancois: From a developer perspective, I'd love if this wouldn't become `null` in special cases. Could we keep it not null and simply return empty when calling `modelContext.getTools()` instead for instance?

domfarolino: > Could we keep it not null

I tend to agree from a developer predictability standpoint. The trouble is that `ModelContext` is document-scoped, so when you call `navigator.modelContext`, you have to get the return value from from the Window's current document, and in a detached context there is simply no document. So it's not that we're nulling it out, and we can just "keep it not null". It's that in a detached frame, there is no document to grab the right object from.

FWIW, this would only ever happen in the detached document case, where script shouldn't really be running anyways:

```js
const iframe = document.createElement('iframe');
document.body.append(iframe);
const iNavigator = iframe.contentWindow.navigator;
iframe.remove();

iNavigator.modelContext; null.
```

If you pulled out a reference to this object in advance, you could still call `getTools()`, and yes:

 > and simply return empty when calling `modelContext.getTools()` instead

... we'd ensure `getTools()` resolves to an empty list.

----

But if we don't like that, we have another option: go back to making `ModelContext` Navigator-owned, and on each getter call, check to see if the document has changed, and if so discard the old object and instantiate a new `ModelContext` and return it. The upside is the attribute never returns null, but the downside is the value of `navigator.modelContext` can change once throughout its lifetime, so we'd still have to remove `[SameObject]` from the IDL. That's probably not bad, and makes for a decent tradeoff. Maybe we should go that route.

bwalderman: +1 on the latter approach, checking if the document has changed on each getter call so we don't have to return `null`. The only potential issue I can see is a developer gets a reference to the `modelContext` from the previous document and tries to use it in script after the document detaches, but as you've already pointed out, script shouldn't really be running at this point anyway.

bwalderman (changes requested):

index.bs
domfarolino: Hah, well if script shouldn't be running anyways, then we should be clear to return `null` :) But anyways, I think the consistency of never returning null is nice, even if the implementation is a bit more ugly. Done.

index.bs
domfarolino: Thanks, I'll see if it makes sense to do this more (or if I should just give up on adding the fancy code letters in dfn links) outside of the PR!

bwalderman (approved):

Issue #178, opened 2026-05-19
Please formally document Permissions-Policy: tools=() as a security control


DMXMax:
## Summary

The `tools` permissions policy feature is implemented in Chrome 150 and 
confirmed working empirically. When `Permissions-Policy: tools=()` is set 
as an HTTP response header, any call to `navigator.modelContext.registerTool()` 
throws:

`SecurityError: Failed to execute 'registerTool' on 'ModelContext': 
Access to the feature "tools" is disallowed by permissions policy.`

This is a meaningful defensive control for site owners who want to prevent 
WebMCP tool registration on pages where it should never be active — 
authenticated account pages, payment flows, sensitive data pages. 

It is enforced by Chrome above the JavaScript layer, meaning injected 
JavaScript cannot bypass it regardless of load order or shadow attacks.

## The problem

This control is not mentioned anywhere in the public spec, proposal, README, 
or security-privacy.md. Site owners and security teams cannot rely on an 
undocumented behavior that may change without notice.

## Request

1. Formally document `Permissions-Policy: tools=()` in the security 
   considerations section of the spec
2. Clarify whether this is intentional as a security control or incidental 
   web platform practice
3. If intentional, provide guidance on recommended use — specifically for 
   pages that should never expose WebMCP tools

## Reference

Chromium source: `third_party/blink/renderer/core/script_tools/model_context.cc`

```cpp
if (!ExecutionContext::From(script_state)->IsFeatureEnabled(
        network::mojom::PermissionsPolicyFeature::kTools)) {
    exception_state.ThrowSecurityError(kPermissionPolicyNotEnabledError);
```

Tested on Chrome 150.0.0.0 with `#enable-webmcp-testing` flag enabled.

victorhuangwq:
The Chrome contributors will propose the spec changes soon, as per https://github.com/webmachinelearning/webmcp/issues/160

domfarolino:
Yep, this is coming this week!

beaufortfrancois:
@DMXMax https://github.com/webmachinelearning/webmcp/pull/179 has been merged. See https://webmachinelearning.github.io/webmcp/#permissions-policy

domfarolino:
[...]

Issue #117, opened 2026-02-27
Allow iframes to deal with tools in their parent/top-level context


yoavweiss:
As [discussed](https://github.com/webmachinelearning/webmcp/issues/51#issuecomment-3973968414) in #51 , in-page agents may be better off if their are not "in-page" and are able to be spun off to an iframe.

Similarly, it might be good to enable iframes to expose tools to the parent page, so that a single agent can have context from the entire page surface that the user is seeing.

What if we had a way for an iframe to say "I want to register a tool with my parent (or top-level) frame", and the parent had a permission policy that `allow`ed iframes to do that? Wouldn't that be neat? :D

Once #51 settles, it'd be similarly neat to have the iframe be able to execute tools on the parent/top-level, provided similar permissions.

domfarolino:
Agreed! Do you think this is covered by https://github.com/webmachinelearning/webmcp/issues/57 @yoavweiss?

johannhof:
+1 this seems like a dupe of #57 :)

yoavweiss:
#57 is half of it. Once https://github.com/webmachinelearning/webmcp/issues/51#issuecomment-4005445894 lands, exposing that to iframes (with explicit permission from the parent) would be the other half.

Issue #57, opened 2025-11-19
Allow iframes to declare tools


khushalsagar:
The proposal currently restricts only the top-level browsing context to declare tools, see [here](https://github.com/webmachinelearning/webmcp/blob/main/docs/proposal.md#:~:text=Only%20a%20top%2Dlevel%20browsing%20context%2C%20such%20as%20a%20browser%20tab%20can%20be%20a%20model%20context%20provider.): "Only a top-level browsing context, such as a browser tab can be a model context provider."

This means embedded frames can't declare tools, same or cross-origin. This was discussed at the last TPAC, notes [here](https://www.w3.org/2025/11/11-webmachinelearning-minutes.html#39d8).

domfarolino:
Discussed with some folks today, and I agree with the direction the meeting seemed to take towards the end, which is creating a permission policy so the top-level can delegate tool registration to a cross-origin iframe and so on ([note here](https://www.w3.org/2025/11/11-webmachinelearning-minutes.html#39d8:~:text=I%20think%20we%20should%20actually%20go%20with%20a%20permission%20policy%20that%20would%20work%20in%20x%2Dorigin%20context)):

> **johannhof:** I think we should actually go with a permission policy that would work in x-origin context

Note that this would be particularly important to support the MCP Apps "registerTool()" use case that @ochafik is working on in  https://github.com/modelcontextprotocol/ext-apps/pull/72.

The good news is I do think this could be polyfilled in the meantime.

domfarolino:
In particular, we should really try to see if we could build MCP Apps's proposed tool registration & execution mechanism _on top of_ native WebMCP, polyfilling gaps where needed.

johannhof:
I agree that a permissions policy would be a great idea here to protect the embedder from 3P tools while making them usable - we should be a bit mindful of attacks coming *from* the embedder as well though.

Giving embedders the ability to direct the agent to use tools from other origins without those origins also agreeing to their usage might go against the same-origin-policy, though this is uncharted territory of course.

victorhuangwq:
Discussing briefly with @lflores-ms, this could be a good fit for document policy, https://github.com/WICG/document-policy/blob/main/document-policy-explainer.md given that there is a need for two-way opt-in, by both the embedder and embeddee.

johannhof:
Thanks Victor! I agree that this might be a good fit. Would be great to get these security requirements defined in the spec.

Starborn:
[...]

PR #179, opened 2026-05-19
Toolchange event and permissions policy

domfarolino:
This PR contributes to the following issues:
 - https://github.com/webmachinelearning/webmcp/issues/57
 - https://github.com/webmachinelearning/webmcp/issues/117
 - https://github.com/webmachinelearning/webmcp/issues/159
 - https://github.com/webmachinelearning/webmcp/issues/160
 - https://github.com/webmachinelearning/webmcp/issues/178

So far this PR only works on the `registerTool()` API, bringing it up to speed with the web platform tests in https://github.com/web-platform-tests/wpt/tree/master/webmcp/imperative. Subsequent PRs will introduce the `getTools()` and `executeTool()`, to fully close the above issues.

This PR introduces:
 - `registerTool()` changes:
    - The `exposeTo` origin list as a tool registration option
    - Scheme checks for `https` exposed origins; an exception is thrown otherwise
    - Permissions policy check
 - The `toolchange` event name
 - The "notify documents of a tool change" algorithm which fires `toolchange` in all documents in the tree that a tool changes in, in order
 - Renames the "tool calling task source" to "webmcp task source", to reflect the fact that the entire spec will use this not just for tool calling
    - Also adds an example of scheduling guarantees
 - The "tools" permissions policy
 - Notification of `toolchange` in unregistration algorithm.

This PR does not introduce the "*" or "native-agent" keywords that we've proposed, to control a tool's exposure to the built-in agent. That will also be considered in a follow-up.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/179.html" title="Last updated on May 19, 2026, 8:53 PM UTC (42dcc53)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/179/986f461...42dcc53.html" title="Last updated on May 19, 2026, 8:53 PM UTC (42dcc53)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -324,7 +324,7 @@ The {{ModelContext}} interface provides methods for web applications to register
 
 <xmp class="idl">
 [Exposed=Window, SecureContext]
-interface ModelContext {
+interface ModelContext : EventTarget {
   undefined registerTool(ModelContextTool tool, optional ModelContextRegisterToolOptions options = {});
 
   attribute EventHandler ontoolchange;
--- a/index.bs
+++ b/index.bs
@@ -101,6 +101,9 @@ spec: html; type: dfn; urlPrefix: https://html.spec.whatwg.org/C
   # `url`.
   text:is initial about:blank; url: is-initial-about:blank
   text: associated Navigator
+spec: SECURE-CONTEXTS; urlPrefix: https://w3c.github.io/webappsec-secure-contexts/
+  type: abstract-op
+    text: is origin potentially trustworthy?; url: is-origin-trustworthy
 </pre>
 
 <h2 id="intro">Introduction</h2>
@@ -402,8 +405,9 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
 
           1. Let |parsedURL| be the result of running the [=URL parser=] on |origin|.
 
-          1. If |parsedURL| is failure or its [=url/scheme=] is not "<code>https</code>", then
-             [=exception/throw=] an "{{SecurityError}}" {{DOMException}}.
+          1. If |parsedURL| is failure or its [=url/origin=] is not [$is origin potentially
+             trustworthy?|potentially trustworthy$], then [=exception/throw=] an "{{SecurityError}}"
+             {{DOMException}}.
 
           1. [=list/Append=] |parsedURL|'s [=url/origin=] to |exposed origins|.
 
--- a/index.bs
+++ b/index.bs
@@ -609,8 +609,7 @@ loop=], that get registered with the APIs in this specification.
 
 The [=user agent=]'s [=browser agent=] runs [=in parallel=] to any [=event loops=] associated
 with a {{ModelContext}} [=relevant global object=]. Steps running on the [=browser agent=] get
-queued on its <dfn>webmcp parallel queue</dfn>, which is the result of [=starting a new parallel
-queue=].
+queued on its <dfn>AI agent queue</dfn>, which is the result of [=starting a new parallel queue=].
 
 Conversely, steps queued *from* the [=browser agent=] onto the [=event loop=] of a given
 {{ModelContext}} object (i.e., the "main thread" where JavaScript runs) are queued on its [=relevant
@@ -646,7 +645,7 @@ in the Chromium project for an example of what might contribute to an observatio
 To <dfn>perform an observation</dfn> given a [=top-level traversable=] |traversable|, run these
 steps:
 
-1. [=Assert=]: This algorithm is running in the [=browser agent=]'s [=webmcp parallel queue=].
+1. [=Assert=]: This algorithm is running in the [=browser agent=]'s [=AI agent queue=].
 
 1. [=Assert=]: |traversable|'s [=navigable/active document=] is not [=Document/fully active=].
 
@@ -691,10 +690,10 @@ Each {{Document}} object has a <dfn for=Document>unique ID</dfn>, which is a [=u
 value=].
 
 The times at which a [=browser agent=] [=performs an observation=] are [=implementation-defined=].
-A [=browser agent=] may [=parallel queue/enqueue steps=] to the [=webmcp parallel queue=] to
-[=perform an observation=] given any [=top-level browsing context=] in the [=user agent=] [=browsing
-context group set=], at any time, although implementations typically reserve this operation for when
-the user is interacting with a [=browser agent=] while web content is in view.
+A [=browser agent=] may [=parallel queue/enqueue steps=] to the [=AI agent queue=] to [=perform an
+observation=] given any [=top-level browsing context=] in the [=user agent=] [=browsing context
+group set=], at any time, although implementations typically reserve this operation for when the
+user is interacting with a [=browser agent=] while web content is in view.
 
 
 <h2 id="security-privacy">Security and privacy considerations</h2>
--- a/index.bs
+++ b/index.bs
@@ -78,6 +78,13 @@ p + dl.props { margin-top: -0.5em; }
 [data-algorithm] [data-algorithm] {
   margin: 1em 0;
 }
+
+/* Table styling definitions from https://resources.whatwg.org/standard.css */
+table { border-collapse: collapse; border-style: hidden hidden none hidden; margin: 1.25em 0; }
+table thead, table tbody { border-bottom: solid; }
+table tbody th { text-align: left; }
+table tbody th:first-child { border-left: solid; }
+table td, table th { border-left: solid; border-right: solid; border-bottom: solid thin; vertical-align: top; padding: 0.2em; }
 </style>
 
 <pre class="link-defaults">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -244,6 +326,8 @@ The {{ModelContext}} interface provides methods for web applications to register
 [Exposed=Window, SecureContext]
 interface ModelContext {
```
```suggestion
interface ModelContext : EventTarget {

File: index.bs [hunk 2]
```diff
@@ -265,6 +349,15 @@ is a [=model context=] [=struct=] created alongside the {{ModelContext}}.
 <div algorithm>
 The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var>)</dfn> method steps are:
 
+1. Let |toolOwner| be [=this=]'s [=relevant global object=]'s [=associated Document|associated
```
beaufortfrancois: Is it possible that `|toolOwner|` is null?

index.bs
```diff
@@ -265,6 +349,15 @@ is a [=model context=] [=struct=] created alongside the {{ModelContext}}.
...
+   <code>Document</code>=].
+
+1. If |toolOwner| is not [=Document/fully active=], then [=exception/throw=] an
+   "{{InvalidStateError}}" {{DOMException}}.
+
+1. If |toolOwner| is not [=allowed to use=] the "{{tools}}" feature, then [=exception/throw=] a
```
beaufortfrancois: Chromium currently throws a security error: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/script_tools/model_context.cc;l=277;drc=a69e289e545ef214a2050fa940f9e4d62e5d8a98

Which one should we pick?

index.bs
```diff
@@ -301,6 +394,19 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
 
+1. Let |exposed origins| be an empty [=list=] of [=origins=].
+
+1. If |options|'s {{ModelContextRegisterToolOptions/exposedTo}} [=map/exists=], then:
+
+     1. [=list/For each=] |origin| of |options|'s {{ModelContextRegisterToolOptions/exposedTo}}:
+
+          1. Let |parsedURL| be the result of running the [=URL parser=] on |origin|.
+
+          1. If |parsedURL| is failure or its [=url/scheme=] is not "<code>https</code>", then
```
beaufortfrancois: How about we allow secure contexts instead of https:// URLs so that localhost works as well to help developers during development?

index.bs
```diff
@@ -168,18 +175,93 @@ A <dfn>tool definition</dfn> is a [=struct=] with the following [=struct/items=]
...
+   navigable=]'s [=Document/descendant navigables=].
+
+1. [=list/For each=] |navigable| of |navigablesToNotify|:
+
+     1. Let |targetDocument| be |navigable|'s [=navigable/active document=].
+
+     1. If |targetDocument| is not [=allowed to use=] the "{{tools}}" feature, then
+        [=iteration/continue=].
+
+     1. If [=tool is visible to an origin=] given |exposed origins| and |targetDocument|'s
```
beaufortfrancois: It's not clear how when this array is empty or omitted, the tool is exposed to documents that are same-origin with the one that registered the tool.

index.bs
```diff
@@ -168,18 +175,93 @@ A <dfn>tool definition</dfn> is a [=struct=] with the following [=struct/items=]
...
+
+<div algorithm>
+To determine if a <dfn>tool is visible to an origin</dfn> given a [=list=] of
+[=origins=] |exposed origins| and an [=origin=] |targetOrigin|, run these steps:
+
+1. [=list/For each=] |origin| of |exposed origins|:
+
+     1. If |targetOrigin| is [=same origin=] with |origin|, then return true.
+
+1. Return false.
```
beaufortfrancois, line 247: I may misunderstand it but it looks like if `|exposed origins|` is empty, it will always return false, which is the case for all registerTool calls without exposedTo. Is that intended?

index.bs
```diff
@@ -168,18 +175,93 @@ A <dfn>tool definition</dfn> is a [=struct=] with the following [=struct/items=]
...
+
+  <xmp class=language-js>
+  navigator.modelContext.ontoolchange = e => console.log('Parent toolchange');
+  iframe.contentWindow.navigator.modelContext.ontoolchange = e => console.log('Child toolchange');
+
+  // Queues a task to fire `toolchange`, on the `webmcp task source`.
+  navigator.modelContext.registerTool({
+    name: "tool_name",
+    description: "tool_desc",
+    execute: async () => {}
```
beaufortfrancois, line 224: Nit: `async` is not mandatory

beaufortfrancois (changes requested):

index.bs
beaufortfrancois: Yep. It looks like it is missing this line: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/model_context_user_data.cc;l=25;drc=59ec7148de9f5edd142cce9c2ffb19793f55e501

index.bs
domfarolino: No, the relevant global (`Window)` and [associated Document](https://html.spec.whatwg.org/#concept-document-window) are never non-null, so the full active check below on this document is the right way to protect against both the detached document case and the bf-cache case. The document's [browsing context could be](https://html.spec.whatwg.org/#window-bc) can be null in the detached document case, but it's not necessary here, since we have the broader fully active check.

The reason this differs from the Chromium implementation is because there, platform objects like `ModelContext` and `Navigator` can have a null Window/Document when constructed in the detached document case, and Chromium allocates the `ModelContext` lazily, which can happen possible after detachment.

index.bs
domfarolino: NotAllowed is more appropriate here, so the spec should stay as is, and I'm updating the implementation to match. Standby for a Chromium change.

index.bs
domfarolino: Trueeeeee, I just wanted to drag into this example the fact that this is Promise-returning, even if it doesn't have to be explicitly.

index.bs
domfarolino: Great catch, and I've been meaning to do this in Chromium too, per https://issues.chromium.org/issues/509983801. Done.

index.bs
domfarolino: Thanks for catching this! Done.

index.bs
domfarolino: Well per your other comment, now it returns true in the same-origin case by default, which is the intention, until we introduce a keyword like `agent`, `native-agent`, or `*` or something. So I think this is good?

index.bs
```diff
@@ -316,7 +414,21 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
+
+1. Let |exposed origins| be an empty [=list=] of [=origins=].
+
+1. If |options|'s {{ModelContextRegisterToolOptions/exposedTo}} [=map/exists=], then:
+
+     1. [=list/For each=] |origin| of |options|'s {{ModelContextRegisterToolOptions/exposedTo}}:
+
+          1. Let |parsedURL| be the result of running the [=URL parser=] on |origin|.
+
+          1. If |parsedURL| is failure or its [=url/origin=] is not [$is origin potentially
```
beaufortfrancois, line 427: It looks like failure is not handled in https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/script_tools/model_context.cc;l=342;drc=a69e289e545ef214a2050fa940f9e4d62e5d8a98. Am I right?
Should we check `!origin`?

```diff
-      if (origin->Protocol() != "https") {
+      if (!origin || origin->Protocol() != "https") {
        exception_state.ThrowSecurityError(
            "Only HTTPS origins are allowed in exposedTo list.");
        return;
      }

beaufortfrancois (approved):

index.bs
domfarolino: It's subtle in Chromium, but that method will never return `nullptr`, and `SecurityOrigin` doesn't have a `operator boolean` override, so I don't think that would do anything. Instead, the string will fail to parse as a URL first, giving an "invalid URL" (equivalent to spec "failure"), and [SecurityOrigin will process it as opaque](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/weborigin/security_origin.cc;l=98-100;drc=3a644a3bda5c62358d9af9bc0705e049d3017375), and we'll fail the below check all the same way.

All of this is handled in CL I'm putting up to make Chrome's behavior match the subtleties in this spec PR that deviate from Chrome—sorry, I've mostly been tracking those in my head and getting them down in a CL, but left the details out of this PR since it's implementation-specific. But perhaps I can be more explicit about communicating when something doesn't match here, and what my plans are to handle it in Chrome.

domfarolino:
Since this PR matches the Chromium behavior and web platform tests I wrote, I'll go ahead and land this. Thanks for the review @beaufortfrancois.

PR #183, opened 2026-05-22
Explainer updates/rewrite

domfarolino:
The explainer is LONG overdue for an update. It's outdated, longwinded, and overly-speculative, and could use some re-organization. We knew this, and we also got this feedback when Chromium was putting the API up for an Origin Trial.

This proposal slims down the explainer (`README.md`), and removes `docs/explainer.md` and `docs/proposal.md`, folding in all of the API design specifics into the main readme. It generally makes all of the prose in the explainer much more concise, and leaves anything roughly speculative to a new and improved "Open questions" section. There are some sections from the original readme that got outright cut:

```md
### Security considerations

### Model poisoning

### Cross-Origin Isolation

### Permissions
```

...because I found that they were mostly redundant with:
 - Things in the `./docs/security-privacy-considerations.md` explainer, which is still linked to from the main readme:
 - Parts of https://github.com/webmachinelearning/webmcp/pull/181
 - Much of the rest of the newly-refactored explainer

The declarative API explainer is still separate, but linked to from the detailed design discussion.

Also much of the https://github.com/webmachinelearning/webmcp#model-context-protocol-mcp-without-webmcp and https://github.com/webmachinelearning/webmcp/blob/main/docs/proposal.md#intersection-with-mcp sections didn't make it over since they spoke much more as if WebMCP was nearly directly usable by MCP-supporting clients, but with discussions about streamable input/outputs, including [transferable objects](https://html.spec.whatwg.org/#transferable-objects) like `ReadableStream`s and specific web platform APIs for user elicitation, it's not clear to me that we should portray ourselves as keeping in lock-step with MCP anymore than the basics.


bwalderman (approved):
LGTM. I have a PR #185 to update the diagrams as well. Recommend merging that first if it looks good to you.

PR #184, opened 2026-05-26
Move the `modelContext` getter to `Document`

### LINKED ISSUE #173: Make tools Document-scoped instead of Window-scoped

domfarolino:
Closes https://github.com/webmachinelearning/webmcp/issues/173.

This PR does not update the explainer or associated examples. I'll make sure https://github.com/webmachinelearning/webmcp/pull/183 does this, since it handles the explainer overhaul.

/cc @beaufortfrancois

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/184.html" title="Last updated on May 27, 2026, 3:04 PM UTC (a33b7cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/184/a816d2e...a33b7cb.html" title="Last updated on May 27, 2026, 3:04 PM UTC (a33b7cb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -100,7 +100,6 @@ spec: html; type: dfn; urlPrefix: https://html.spec.whatwg.org/C
   # exclude `url`. But the colon breaks the URL in `text`, forcing us to include
   # `url`.
   text:is initial about:blank; url: is-initial-about:blank
-  text: associated Navigator
 spec: SECURE-CONTEXTS; urlPrefix: https://w3c.github.io/webappsec-secure-contexts/
   type: abstract-op
     text: is origin potentially trustworthy?; url: is-origin-trustworthy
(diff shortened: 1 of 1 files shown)


bwalderman (approved):

index.bs
```diff
@@ -100,7 +100,6 @@ spec: html; type: dfn; urlPrefix: https://html.spec.whatwg.org/C
   # exclude `url`. But the colon breaks the URL in `text`, forcing us to include
   # `url`.
   text:is initial about:blank; url: is-initial-about:blank
-  text: associated Navigator
```
beaufortfrancois: I think we can remove all of this as well right?

beaufortfrancois (approved):
LGTM with nit

anowardear062-svg (approved):

index.bs
domfarolino: Ah yeah, nice catch.

Issue #193, opened 2026-05-28
Self-Review Questionnaire: Security and Privacy


anssiko:
Opening this issue to track the self-review questionnaire part of this resolution:

>[RESOLUTION](https://www.w3.org/2026/05/28-webmachinelearning-minutes.html#a63a): The group uses the updated Explainer and the newly authored Privacy and Security Considerations as a reference, and adds a self-review questionnaire responses, then stages review requests for the group to review before submission.

Thank you @victorhuangwq for volunteering to lead this.

Self-Review Questionnaire: Security and Privacy
https://w3c.github.io/security-questionnaire/

victorhuangwq:
Thanks, I will get started on the questionnaire in a draft PR. Will mark it ready for review once ready. 

I will use this thread for anything that I require input from the community as well.

domfarolino:
Closing this since #195 landed.

PR #195, opened 2026-05-29
Add security and privacy questionnaire

victorhuangwq:
Addresses https://github.com/webmachinelearning/webmcp/issues/193

cc: @bwalderman @johannhof @domfarolino

### DIFF
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -2,7 +2,7 @@
 
 > 01. What information does this feature expose, and for what purposes?
 
-WebMCP exposes author-defined tool metadata and tool return values to the user agent. It does not expose new information about the user or their environment to origins.
+WebMCP exposes author-defined tool metadata and tool return values to the built-in AI agent. It does not expose new information about the user or their environment to origins.
 
 Cross-origin iframes may discover these tools only if the tool author explicitly opts in via `exposedTo`.
 
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -74,7 +74,7 @@ Additionally, tools can specify [`exposedTo`](https://webmachinelearning.github.
 
 > 15. How do the features in this specification work in the context of a browser's Private Browsing or Incognito mode?
 
-We do not anticipate any differences.
+We do not anticipate any differences, but implementers should be aware of how to safely handle private browsing modes. See [Interaction with Private Browsing Modes](https://webmachinelearning.github.io/webmcp/#interaction-with-private-browsing).
 
 > 16. Does this specification have both "Security Considerations" and "Privacy Considerations" sections?
 
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -14,7 +14,7 @@ Yes. The API surface exposes only what is necessary for agents to discover and i
 
 No, the API itself does not expose PII, but the tools that authors choose to implement _can_, depending on their nature.
 
-We note a novel challenge for agent implementers: malicious tools can request a non-minimal set of personal data via their input parameters, causing privacy leakage. See [Privacy Leakage through Over-Parameterization](https://w3c.github.io/webmcp/#privacy-leakage-over-parameterization) for details. WebMCP does not increase the attack vector compared to tools in non-WebMCP contexts, but agent implementers should be aware that this risk exists.
+We note a novel challenge for agent implementers: malicious tools can request a non-minimal set of personal data via their input parameters, causing privacy leakage. See [Privacy Leakage through Over-Parameterization](https://webmachinelearning.github.io/webmcp/#privacy-leakage-over-parameterization) for details. WebMCP does not increase the attack vector compared to tools in non-WebMCP contexts, but agent implementers should be aware that this risk exists.
 
 > 04. How do the features in your specification deal with sensitive information?
 
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -107,6 +107,6 @@ No.
 
 > 22. What should this questionnaire have asked?
 
-We would like to surface a risk not directly surfaced by this questionnaire: agents browsing multiple origins may carry state from one origin to another, potentially leading to data leakage or same-origin policy bypasses if not handled securely by the user agent. This risk is inherent to multi-origin agent browsing and exists independently of WebMCP.
+We would like to surface a risk not directly surfaced by this questionnaire: agents browsing multiple origins may carry state from one origin to another if not handled securely by the user agent. This risk is inherent to multi-origin agent browsing and exists independently of WebMCP.
 
 This section is currently marked as TODO in the draft spec, and we intend to update it with our analysis soon. See [Violation of Same-Origin Boundaries](https://webmachinelearning.github.io/webmcp/#violation-same-origin-boundaries).
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -32,7 +32,7 @@ No. Tool registrations are tied to the document's lifetime. There are discussion
 
 > 07. Do the features in your specification expose information about the underlying platform to origins?
 
-No.
+No. While the API introduces a communication channel to agents that could have information about the underlying platform, these agents were already able to provide this information to the site through other channels, e.g. local network endpoints or filling HTML forms.
 
 > 08. Does this specification allow an origin to send data to the underlying platform?
 
@@ -40,7 +40,7 @@ Yes, tool inputs and outputs flow between an origin and the platform's built-in 
 
 > 09. Do features in this specification enable access to device sensors?
 
-No.
+No. While the API introduces a communication channel to agents that could have access to device sensors, these agents were already able to communicate with the site through other channels, e.g. local network endpoints or filling HTML forms.
 
 > 10. Do features in this specification enable new script execution/loading mechanisms?
 
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -44,7 +44,9 @@ No.
 
 > 10. Do features in this specification enable new script execution/loading mechanisms?
 
-No. Tool `execute` callbacks are ordinary JavaScript invoked in the registering document's existing realm.
+Yes, it introduces a new script invocation mechanism. Agents (and, when the author opts in via [`exposedTo`](https://webmachinelearning.github.io/webmcp/#dom-modelcontextregistertooloptions-exposedto), authorized cross-origin documents) can directly invoke a tool's [`execute`](https://webmachinelearning.github.io/webmcp/#dom-modelcontexttool-execute) callback with structured, schema-conforming arguments. 
+
+These callbacks are ordinary JavaScript running in the registering document's existing realm, no new script content can be loaded.
 
 > 11. Do features in this specification allow an origin to access other devices?
 
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -107,4 +107,6 @@ No.
 
 > 22. What should this questionnaire have asked?
 
-None that we can think of.
+We would like to bring attention to a security risk not directly surfaced by this questionnaire: agents browsing multiple origins may carry state from one origin to another, potentially leading to data leakage or same-origin policy bypasses if not handled securely by the user agent. This risk is inherent to multi-origin agent browsing and exists independently of WebMCP.
+
+This section is currently marked as TODO in the draft spec, and we intend to update it with our analysis soon. See [Violation of Same-Origin Boundaries](https://webmachinelearning.github.io/webmcp/#violation-same-origin-boundaries).
--- a/security-privacy-questionnaire.md
+++ b/security-privacy-questionnaire.md
@@ -1,95 +1,95 @@
 # [Self-Review Questionnaire: Security and Privacy](https://w3c.github.io/security-questionnaire/)
 
-01.  What information does this feature expose, and for what purposes?
+> 01. What information does this feature expose, and for what purposes?
 
 WebMCP exposes author-defined tool metadata and tool return values to the user agent. It does not expose new information about the user or their environment to origins.
 
-Cross-origin iframes may discover these tools only if the tool author explicitly opts in via `exposedTo`. 
+Cross-origin iframes may discover these tools only if the tool author explicitly opts in via `exposedTo`.
 
-02.  Do features in your specification expose the minimum amount of information necessary to implement the intended functionality?
+> 02. Do features in your specification expose the minimum amount of information necessary to implement the intended functionality?
 
-Yes. The API surface exposes only what is necessary for agents to discover and invoke tools. The information that flows through tool parameters and return values is entirely scoped to what the author declares. 
+Yes. The API surface exposes only what is necessary for agents to discover and invoke tools. The information that flows through tool parameters and return values is entirely scoped to what the author declares.
 
-03.  Do the features in your specification expose personal information, personally-identifiable information (PII), or information derived from either?
+> 03. Do the features in your specification expose personal information, personally-identifiable information (PII), or information derived from either?
 
 No, the API itself does not expose PII.
 
-We note a novel challenge for agent implementors: malicious tools can request a non-minimal set of personal data via their input parameters, causing privacy leakage. See [Privacy Leakage through Over-Parameterization](https://w3c.github.io/webmcp/#privacy-leakage-over-parameterization) for details. WebMCP does not increase the attack vector compared to tools in non-WebMCP contexts, but implementors should be aware that this risk exists. 
+We note a novel challenge for agent implementors: malicious tools can request a non-minimal set of personal data via their input parameters, causing privacy leakage. See [Privacy Leakage through Over-Parameterization](https://w3c.github.io/webmcp/#privacy-leakage-over-parameterization) for details. WebMCP does not increase the attack vector compared to tools in non-WebMCP contexts, but implementors should be aware that this risk exists.
 
-04.  How do the features in your specification deal with sensitive information?
+> 04. How do the features in your specification deal with sensitive information?
 
-WebMCP is not a source of sensitive information. Tools may wrap sensitive or high-privilege operations (e.g., purchases, account changes), but that is not WebMCP-specific. We discuss this risk in [Tool Implementation as Attack Targets](https://webmachinelearning.github.io/webmcp/#tool-implementation-targets).
+WebMCP is not a source of sensitive information. Tools may wrap sensitive or high-privilege operations (e.g., purchases, account changes), but that risk is not WebMCP-specific. We discuss this risk in [Tool Implementation as Attack Targets](https://webmachinelearning.github.io/webmcp/#tool-implementation-targets).
 
-05.  Does data exposed by your specification carry related but distinct information that may not be obvious to users?
+> 05. Does data exposed by your specification carry related but distinct information that may not be obvious to users?
 
 No, the API surface itself does not carry related but distinct information.
 
-06.  Do the features in your specification introduce state that persists across browsing sessions?
+> 06. Do the features in your specification introduce state that persists across browsing sessions?
 
 No. Tool registrations are tied to the document's lifetime. There are discussions about persisting tools across navigation, but that is not currently specified.
 
-07.  Do the features in your specification expose information about the underlying platform to origins?
+> 07. Do the features in your specification expose information about the underlying platform to origins?
 
 No.
 
-08.  Does this specification allow an origin to send data to the underlying platform?
+> 08. Does this specification allow an origin to send data to the underlying platform?
 
 Tool inputs and outputs flow between the page and the authorized agent, which may include the user agent's built-in agent. The data is structured (JSON-serializable values conforming to declared schemas).
 
-09.  Do features in this specification enable access to device sensors?
+> 09. Do features in this specification enable access to device sensors?
 
 No.
 
-10.  Do features in this specification enable new script execution/loading mechanisms?
+> 10. Do features in this specification enable new script execution/loading mechanisms?
 
 No. Tool `execute` callbacks are ordinary JavaScript invoked in the registering document's existing realm.
 
-11.  Do features in this specification allow an origin to access other devices?
+> 11. Do features in this specification allow an origin to access other devices?
 
 No.
 
-12.  Do features in this specification allow an origin some measure of control over a user agent's native UI?
+> 12. Do features in this specification allow an origin some measure of control over a user agent's native UI?
 
 No direct control. There is discussion of `requestUserInput` in [Issue #165](https://github.com/webmachinelearning/webmcp/issues/165).
 
-13.  What temporary identifiers do the features in this specification create or expose to the web?
+> 13. What temporary identifiers do the features in this specification create or expose to the web?
 
 None.
 
-14.  How does this specification distinguish between behavior in first-party and third-party contexts?
+> 14. How does this specification distinguish between behavior in first-party and third-party contexts?
 
 The feature is gated by Permissions Policy `"tools"`. It is allowed in top-level documents and same-origin descendants by default; Permissions Policy can be used to allow it in cross-origin iframes and/or to disallow it in same-origin frames.
 
 Additionally, tools can specify `exposedTo` to control which origins (or `native-agents`, name to be bikeshed per [#179](https://github.com/webmachinelearning/webmcp/pull/179)) can discover them.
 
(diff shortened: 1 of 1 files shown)


security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
+# [Self-Review Questionnaire: Security and Privacy](https://w3c.github.io/security-questionnaire/)
+
+> 01. What information does this feature expose, and for what purposes?
+
+WebMCP exposes author-defined tool metadata and tool return values to the user agent. It does not expose new information about the user or their environment to origins.
+
+Cross-origin iframes may discover these tools only if the tool author explicitly opts in via `exposedTo`.
```
domfarolino: Link to https://webmachinelearning.github.io/webmcp/#dom-modelcontextregistertooloptions-exposedto

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
+# [Self-Review Questionnaire: Security and Privacy](https://w3c.github.io/security-questionnaire/)
+
+> 01. What information does this feature expose, and for what purposes?
+
+WebMCP exposes author-defined tool metadata and tool return values to the user agent. It does not expose new information about the user or their environment to origins.
```
domfarolino:
```suggestion
WebMCP exposes author-defined tool metadata and tool return values to the built-in AI agent. It does not expose new information about the user or their environment to origins.
```

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
...
+
+Cross-origin iframes may discover these tools only if the tool author explicitly opts in via `exposedTo`.
+
+> 02. Do features in your specification expose the minimum amount of information necessary to implement the intended functionality?
+
+Yes. The API surface exposes only what is necessary for agents to discover and invoke tools. The information that flows through tool parameters and return values is entirely scoped to what the author declares.
```
domfarolino:
```suggestion
Yes. The API surface exposes only what is necessary for agents to discover and invoke tools. The information that flows through tool metadata like parameters and annotations, as well as tool return values, is entirely scoped to what the author declares.
```

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
...
+
+> 03. Do the features in your specification expose personal information, personally-identifiable information (PII), or information derived from either?
+
+No, the API itself does not expose PII.
```
domfarolino:
```suggestion
No, the API itself does not expose PII, but the tools that authors choose to implement _can_, depending on their nature.
```

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
...
+
+We note a novel challenge for agent implementors: malicious tools can request a non-minimal set of personal data via their input parameters, causing privacy leakage. See [Privacy Leakage through Over-Parameterization](https://w3c.github.io/webmcp/#privacy-leakage-over-parameterization) for details. WebMCP does not increase the attack vector compared to tools in non-WebMCP contexts, but implementors should be aware that this risk exists.
```
domfarolino:
```suggestion
We note a novel challenge for agent implementers: malicious tools can request a non-minimal set of personal data via their input parameters, causing privacy leakage. See [Privacy Leakage through Over-Parameterization](https://w3c.github.io/webmcp/#privacy-leakage-over-parameterization) for details. WebMCP does not increase the attack vector compared to tools in non-WebMCP contexts, but agent implementers should be aware that this risk exists.
```

small nit: https://whatwg.org/style-guide#dictionary

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
...
+
+> 04. How do the features in your specification deal with sensitive information?
+
+WebMCP is not a source of sensitive information. Tools may wrap sensitive or high-privilege operations (e.g., purchases, account changes), but that risk is not WebMCP-specific. We discuss this risk in [Tool Implementation as Attack Targets](https://webmachinelearning.github.io/webmcp/#tool-implementation-targets).
```
domfarolino, line 21: Well said!

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
...
+
+No. Tool registrations are tied to the document's lifetime. There are discussions about persisting tools across navigation, but that is not currently specified.
+
+> 07. Do the features in your specification expose information about the underlying platform to origins?
+
+No.
+
+> 08. Does this specification allow an origin to send data to the underlying platform?
+
+Tool inputs and outputs flow between the page and the authorized agent, which may include the user agent's built-in agent. The data is structured (JSON-serializable values conforming to declared schemas).
```
domfarolino:
```suggestion
Yes, tool inputs and outputs flow between an origin and the platform's built-in agent. The data is structured JSON-serializable values conforming to declared schemas.
```

To make it more explicit, regarding the part of WebMCP that is aimed at built-in agents. I think this is a slightly more straightforward answer.

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
...
+
+> 09. Do features in this specification enable access to device sensors?
+
+No.
+
+> 10. Do features in this specification enable new script execution/loading mechanisms?
+
+No. Tool `execute` callbacks are ordinary JavaScript invoked in the registering document's existing realm.
```
domfarolino:
```suggestion
No. Tool [`execute`](https://webmachinelearning.github.io/webmcp/#dom-modelcontexttool-execute) callbacks are ordinary JavaScript invoked in the registering document's existing realm.
```

If we keep this answer, I have the suggestion above. However, I think there's a chance we might want to say "Yes" to this question, since we're basically introducing a structured postMessage() v2, which allows a set of authorized origins to basically directly schedule callbacks in the tool provider's realm, which is kind of "new" on the platform. Since this is relevant to the TAG review we'll file, let me just ask @jyasskin directly: what do you think our answer should be for this question?

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
...
+
+No direct control. There is discussion of `requestUserInput` in [Issue #165](https://github.com/webmachinelearning/webmcp/issues/165).
+
+> 13. What temporary identifiers do the features in this specification create or expose to the web?
+
+None.
+
+> 14. How does this specification distinguish between behavior in first-party and third-party contexts?
+
+The feature is gated by Permissions Policy `"tools"`. It is allowed in top-level documents and same-origin descendants by default; Permissions Policy can be used to allow it in cross-origin iframes and/or to disallow it in same-origin frames.
```
domfarolino:
```suggestion
The feature is gated by the [`"tools"`](https://webmachinelearning.github.io/webmcp/#permissiondef-tools) permission policy. It is allowed in top-level documents and same-origin descendants by default; The permission policy can be used to allow it in cross-origin iframes and/or to disallow it in same-origin frames.
```

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
...
+
+Additionally, tools can specify `exposedTo` to control which origins (or `native-agents`, name to be bikeshed per [#179](https://github.com/webmachinelearning/webmcp/pull/179)) can discover them.
```
domfarolino: Link to exposedTo again.

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
...
+
+Yes. See [Security and Privacy Considerations](https://webmachinelearning.github.io/webmcp/#security-privacy).
+
+> 17. Do features in your specification enable origins to downgrade default security protections?
+
+No.
+
+> 18. What happens when a document that uses your feature is kept alive in BFCache (instead of getting destroyed) after navigation, and potentially gets reused on future navigations back to the document?
+
+A BFCached document's registered tools remain in memory. While the document is non-fully-active, agents cannot invoke its tools or deliver events to it. On restoration, registered tools become available again.
```
domfarolino: Also add a note about how a document cannot register tools while in the bfcache, nor can it retrieve the list of tools available to it.

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
...
+
+> 19. What happens when a document that uses your feature gets disconnected?
+
+A disconnected document's tools are no longer discoverable or invokable by agents. Pending tool invocations associated with the document are abandoned.
```
domfarolino: This is not spec'd yet, so let's make a note that this will be the case, but the spec needs to catch up. And let's clarify that "abandoned" means that:
 - In-page agents: the caller's Promise will be rejected
 - Built-in agents: the agent will be notified that the tool call failed.

security-privacy-questionnaire.md
bvandersloot-mozilla: FWIW, I'd argue in favor of "yes" here.

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,95 @@
...
+
+No. Tool `execute` callbacks are ordinary JavaScript invoked in the registering document's existing realm.
+
+> 11. Do features in this specification allow an origin to access other devices?
+
+No.
+
+> 12. Do features in this specification allow an origin some measure of control over a user agent's native UI?
+
+No direct control. There is discussion of `requestUserInput` in [Issue #165](https://github.com/webmachinelearning/webmcp/issues/165).
```
bvandersloot-mozilla: Shouldn't it be "yes"? Even the readOnlyHint offers indirect control over the browsing agent's native UI.

security-privacy-questionnaire.md
bvandersloot-mozilla, line 21: One question I anticipate from privacy review is what mitigations for this issue are possible for WebMCP. I agree with Dom: well said to make clear WebMCP isn't creating the problem. However, if there are any locations to tie the agent's hands a bit and make the privacy story better (even on the level of recommendations rather than requirements) that would be an improvement. Alternatively, you should make clear that nothing is actually being done to defend against that attack in the spec with something like `... but that risk is not WebMCP-specific and is therefore not defended against in the design of the API.`

security-privacy-questionnaire.md
victorhuangwq: Updated!

security-privacy-questionnaire.md
victorhuangwq: Good point, I initially viewed readOnlyHint and similar annotations as purely advisory, but you're right that they do give origins indirect influence over how the agent presents tool invocations in its UI. I've updated the answer to acknowledge this. Thanks!

security-privacy-questionnaire.md
victorhuangwq: I think we can reference the https://webmachinelearning.github.io/webmcp/#mitigations section to begin with perhaps? We should also update the mitigations to include this https://github.com/webmachinelearning/webmcp/issues/176, as discussed in the CG call.

Beyond that, there will just be things that will be left as recommendations for agent implementers, and thus non-normative in nature. But there should be areas where the API design can help with, and I don't want to close the doors on future mitigations by stating otherwise

security-privacy-questionnaire.md
victorhuangwq: Make sense

security-privacy-questionnaire.md
victorhuangwq: Suggestion taken - I think it does preempt predictable reviewer question. Though the exposure is akin to developers displaying PII on the account settings page.

security-privacy-questionnaire.md
johannhof: +1 to Victor here, strong normative protection might be difficult, but we should definitely try to add e.g. hints that help the agent manage risky actions.

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,102 @@
...
+
+No, the API surface itself does not carry related but distinct information.
+
+> 06. Do the features in your specification introduce state that persists across browsing sessions?
+
+No. Tool registrations are tied to the document's lifetime. There are discussions about persisting tools across navigation, but that is not currently specified.
+
+> 07. Do the features in your specification expose information about the underlying platform to origins?
+
+No.
```
johannhof: Well, it could, right? It exposes whatever you want to ask of the agent, if the agent considers it appropriate to expose this information.

So you could write a `do_thing(os_name, os_version, cpu_model)` tool and, say, a local command line agent could easily supply that information.

WebMCP is a "dumb pipe", in that sense.

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,102 @@
...
+
+> 08. Does this specification allow an origin to send data to the underlying platform?
+
+Yes, tool inputs and outputs flow between an origin and the platform's built-in agent. The data is structured JSON-serializable values conforming to declared schemas.
+
+> 09. Do features in this specification enable access to device sensors?
+
+No.
```
johannhof: Like the above, that depends on how it's used. I think it's fair to say "no" here, but we should note the general "pipe-to-powerful-agent" nature of the API.

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,102 @@
...
+
+No. Tool `execute` callbacks are ordinary JavaScript invoked in the registering document's existing realm.
+
+> 11. Do features in this specification allow an origin to access other devices?
+
+No.
+
+> 12. Do features in this specification allow an origin some measure of control over a user agent's native UI?
+
+[Tool annotations](https://webmachinelearning.github.io/webmcp/#dom-modelcontexttoolannotations) can indirectly influence how an agent presents a tool invocation in its UI (e.g., a `readOnlyHint` may cause the agent to skip a confirmation step).
```
johannhof: If the user agent runs a built in agent, the tool responses are presumably also shown and/or influence what you would call "native" UI (usually these things are isolated documents in a sidebar of some sort, but it's generally more privileged than regular websites and may appear part of the "native" experience to users).

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,102 @@
...
+
+> 14. How does this specification distinguish between behavior in first-party and third-party contexts?
+
+The feature is gated by the [`"tools"`](https://webmachinelearning.github.io/webmcp/#permissiondef-tools) permission policy. It is allowed in top-level documents and same-origin descendants by default; The permission policy can be used to allow it in cross-origin iframes and/or to disallow it in same-origin frames.
+
+Additionally, tools can specify [`exposedTo`](https://webmachinelearning.github.io/webmcp/#dom-modelcontextregistertooloptions-exposedto) to control which origins (or `native-agents`, name to be bikeshed per [#179](https://github.com/webmachinelearning/webmcp/pull/179)) can discover them.
+
+> 15. How do the features in this specification work in the context of a browser's Private Browsing or Incognito mode?
+
+We do not anticipate any differences.
```
johannhof: We should probably use this as an opportunity to mention Private / Incognito Mode in the S&P considerations - there is a risk that agents weaken that barrier by not caring about the difference between the two modes and leaking user state into private browsing.

johannhof:
Sorry for the delay, a few comments but L (pretty) GTM already! :)

security-privacy-questionnaire.md
johannhof: Added #201 @victorhuangwq PTAL

security-privacy-questionnaire.md
domfarolino: I think me too. Let's change this to "yes" then.

security-privacy-questionnaire.md
domfarolino: Sounds good, @victorhuangwq do you want to link to the mitigations section in the spec from here then?

security-privacy-questionnaire.md
domfarolino: I guess I'd slightly push back and say in that case, it's not our "features" that are exposing any information about the underlying platform. It's the agent that's exposing it. Just like it might in a standard HTML form _without_ WebMCP. Maybe we should clarify that here. How about:

```
No, WebMCP does not expose information about the underlying platform, but an agent with access to the platform could choose to expose such information through a tool call. However, this exposure is dependent on agent behavior, and is not specifically made possible by WebMCP.
```

security-privacy-questionnaire.md
domfarolino: I think in the spirit of treating the agent like an extension of the user, it would be the equivalent of a pre-AI website asking the user to: _"Input in this HTML form, the result of running the FOO operation on the Bluetooth device plugged into your laptop"_. We wouldn't really say that this site "enables access to device sensors" if the user follows these instructions. I think this question is more reserved for deterministic privilege access that wasn't possible via script before.

security-privacy-questionnaire.md
domfarolino: Since you have a PR for this, I think we can resolve this specific comment.

security-privacy-questionnaire.md
domfarolino: +1. Plus even the tool [title](https://webmachinelearning.github.io/webmcp/#dom-modelcontexttool-title) is exactly designed for this, so it's worth calling out that in addition to outputs.

domfarolino (commented):
Overall it looks like we're pretty close, but there are a few open threads to resolve:

 - https://github.com/webmachinelearning/webmcp/pull/195#discussion_r3374992792
 - https://github.com/webmachinelearning/webmcp/pull/195#discussion_r3375005517
 - https://github.com/webmachinelearning/webmcp/pull/195#discussion_r3375026898
 - https://github.com/webmachinelearning/webmcp/pull/195#discussion_r3375059500
 - https://github.com/webmachinelearning/webmcp/pull/195#discussion_r3373764156

security-privacy-questionnaire.md
johannhof: Something like that seems totally fine, yeah. It's still a machine to machine sort of interface, so it's important to call out any data can be transferred.

security-privacy-questionnaire.md
johannhof: Well the difference is that in this case it can be done semi-automatically without the human explicitly sharing the data, depending on the agent implementation. I generally agree with your sentiment here, but as per the above it should be pointed out that agents can do *anything*.

security-privacy-questionnaire.md
domfarolino: Unresolving since we landed https://github.com/webmachinelearning/webmcp/pull/201 and should probably link to that section of the spec here actually...

security-privacy-questionnaire.md
bvandersloot-mozilla: > It's the agent that's exposing it. Just like it might in a standard HTML form without WebMCP. 

This comes up a lot. Having this written out once in a canonical location in the spec's security and privacy considerations seems like it could be helpful.

security-privacy-questionnaire.md
bvandersloot-mozilla: Linking to mitigations is a fine start.

Just to clear up a couple of misconceptions: 

> Beyond that, there will just be things that will be left as recommendations for agent implementers, and thus non-normative in nature. 

Normative language can include non MUST RFC 2119 terms, like SHOULD and MAY.

> But there should be areas where the API design can help with, and I don't want to close the doors on future mitigations by stating otherwise

Saying that you don't do anything about it is not closing the door on future mitigations, particularly if you say so.

johannhof: IMO those are two fair points from Ben here :)

victorhuangwq: Updated based on feedback.

security-privacy-questionnaire.md
victorhuangwq: I think that make sense - added the clarification that there is ability to do cross origin invocation of tools

security-privacy-questionnaire.md
victorhuangwq: ah, i completely miss the responses and title part, which is a way more direct influence as well. updated.

security-privacy-questionnaire.md
victorhuangwq: rebased and linked to #201

security-privacy-questionnaire.md
victorhuangwq: marking this as resolved

security-privacy-questionnaire.md
johannhof:
```suggestion
No. While the API introduces a communication channel to agents that could have information about the underlying platform, these agents were already able to provide this information to the site through other channels, e.g. local network endpoints or filling HTML forms.
```

security-privacy-questionnaire.md
johannhof:
```suggestion
No. While the API introduces a communication channel to agents that could have access to device sensors, these agents were already able to communicate with the site through other channels, e.g. local network endpoints or filling HTML forms.
```

victorhuangwq:
Just merged @johannhof's suggestion. 

@domfarolino @bvandersloot-mozilla @bwalderman ready for review and merging.

anssiko:
This resolution touches this PR, so filing here:

>[RESOLUTION](https://www.w3.org/2026/06/11-webmachinelearning-minutes.html#3656): The group approves the updated review materials (explainer, S&P considerations, S&P questionnaire). The spec editors will stage the review requests in WebMCP repo for the TAG, Privacy WG and Security WG, ensuring that specific questions or technical issues are highlighted to the reviewers in the appropriate sections of the review requests. (issue [#195](https://github.com/webmachinelearning/webmcp/issues/195))

Once this PR is merged it unblocks staging. FYI @domfarolino @bwalderman

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,110 @@
...
+
+> 02. Do features in your specification expose the minimum amount of information necessary to implement the intended functionality?
+
+Yes. The API surface exposes only what is necessary for agents to discover and invoke tools. The information that flows through tool metadata like parameters and annotations, as well as tool return values, is entirely scoped to what the author declares.
+
+> 03. Do the features in your specification expose personal information, personally-identifiable information (PII), or information derived from either?
+
+No, the API itself does not expose PII, but the tools that authors choose to implement _can_, depending on their nature.
+
+We note a novel challenge for agent implementers: malicious tools can request a non-minimal set of personal data via their input parameters, causing privacy leakage. See [Privacy Leakage through Over-Parameterization](https://w3c.github.io/webmcp/#privacy-leakage-over-parameterization) for details. WebMCP does not increase the attack vector compared to tools in non-WebMCP contexts, but agent implementers should be aware that this risk exists.
```
bwalderman: Fix broken link.

security-privacy-questionnaire.md
bwalderman: It's good that we're stating we'll continue to look at further mitigations but I agree with Ben we should make the currently limitations explicit here. Something like: "The spec does not currently include normative defenses against this; the mitigations listed are recommendations for agent implementers."

security-privacy-questionnaire.md
```diff
@@ -0,0 +1,110 @@
...
+> 20. Does your spec define when and how new kinds of errors should be raised?
+
+Yes. `registerTool()` throws `InvalidStateError` for inactive documents, duplicate names, or invalid name/description; `NotAllowedError` when the `"tools"` Permissions Policy is disallowed; `SecurityError` for non-trustworthy [`exposedTo`](https://webmachinelearning.github.io/webmcp/#dom-modelcontextregistertooloptions-exposedto) origins; and `TypeError` when `inputSchema` serialization fails. These errors only reflect the page's own state and inputs, so they do not leak new information.
+
+> 21. Does your feature allow sites to learn about the user's use of assistive technology?
+
+No.
+
+> 22. What should this questionnaire have asked?
+
```
bwalderman, line 111: There's one more important addition I'd like to make here. 

[Violation of Same-Origin Boundaries](https://webmachinelearning.github.io/webmcp/#violation-same-origin-boundaries) is admittedly still a TODO in our spec but this is a novel security risk that we should raise to reviewers. Our answer at the moment may be the same as our answer about over-parameterization; that this risk in inherent in agents that can browse multiple origins and exists without WebMCP. We should still flag the issue in this questionnaire though to make sure it receives attention.

bwalderman (changes requested):

security-privacy-questionnaire.md
victorhuangwq: Addressed

security-privacy-questionnaire.md
victorhuangwq: addressed as well

security-privacy-questionnaire.md
victorhuangwq: I clarified that that spec does not include normative guidance yet.

bwalderman (approved):

domfarolino (approved):

PR #197, opened 2026-06-05
Gate registerTool on document.domain being disabled

beaufortfrancois:
Following https://chromium-review.googlesource.com/c/chromium/src/+/7896184, this change updates the spec to gate `registerTool()` on `document.domain` being disabled.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/webmcp/pull/197.html" title="Last updated on Jun 5, 2026, 3:27 PM UTC (9065910)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/197/bc9c77a...beaufortfrancois:9065910.html" title="Last updated on Jun 5, 2026, 3:27 PM UTC (9065910)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -332,6 +332,9 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
 1. If |tool owner| is not [=Document/fully active=], then [=exception/throw=] an
    "{{InvalidStateError}}" {{DOMException}}.
 
+1. If [=this=]'s [=relevant global object=]'s {{Window/originAgentCluster}} is false, then
+  [=exception/throw=] a "{{SecurityError}}" {{DOMException}}.
+
 1. If |tool owner| is not [=allowed to use=] the "{{tools}}" feature, then [=exception/throw=] a
    "{{NotAllowedError}}" {{DOMException}}.
 
--- a/index.bs
+++ b/index.bs
@@ -92,11 +92,16 @@ spec:html; type:dfn;
   text:form-associated element
   text:browsing context group set
   text:unique internal value
+  text:is origin-keyed
 </pre>
 <pre class="anchors">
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -332,6 +332,9 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
 1. If |tool owner| is not [=Document/fully active=], then [=exception/throw=] an
    "{{InvalidStateError}}" {{DOMException}}.
 
+1. If [=this=]'s [=relevant global object=]'s {{Window/originAgentCluster}} is false, then
```
domfarolino: This is annoying but we actually can't call the getter directly, since it's possible for script to override the getter, and that would technically trip up this spec prose. So you'll need to write the long-hand:

> If this's [surrounding agent](https://tc39.es/ecma262/#surrounding-agent)'s [agent cluster](https://tc39.es/ecma262/#sec-agent-clusters)'s [is origin-keyed](https://html.spec.whatwg.org/#is-origin-keyed)...

index.bs
```diff
@@ -332,6 +332,9 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
+  [=exception/throw=] a "{{SecurityError}}" {{DOMException}}.
```
domfarolino: We also want an exception for file URLs like Chromium does, right?

index.bs
beaufortfrancois: I didn't know. Thanks!

index.bs
beaufortfrancois: Good catch, I thought it was a Chromium-only implementation thing.

domfarolino (approved):

PR #200, opened 2026-06-08
Make `registerTool()` return a promise

### LINKED ISSUE #175: Consider making `registerTool()` return a Promise
Currently, `registerTool()` expects that tool registration happens synchronously, and all errors related to registration will be thrown synchronously. But with cross-origin iframe integration, tools are shared across the frame tree, and their registration becomes fundamentally asynchronous, even if the registration error conditions are all synchronously known during the call to `registerTool()` (since they only rely on the local, per-document cache of tools).

But since things are becoming async, we might want to consider making `registerTool()` return a Promise. This has two benefits:

1. We could resolve the Promise when the tool is *actually* registered asynchronously in the part of the platform that makes it available to `getTools()` in other documents. This makes the Promise resolution a more "true" representation of the registration process, instead of the caller having to listen to some event in their own document that signals "the list of available tools has changed!"
2. **Forward compatibility**: to keep our options open in the future. Right now we have no asynchronous tool registration failure cases, but what if we want to create some in the future? It's hard to signal async failures ergonomically to a sync API. It'd be a shame if we ever ended up in that case.

Given the above two, I vote that we make `registerTool()` return a Promise.

beaufortfrancois:
This PR makes the `registerTool()` method return a `Promise`; fixes https://github.com/webmachinelearning/webmcp/issues/175.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/webmcp/pull/200.html" title="Last updated on Jun 8, 2026, 4:56 PM UTC (7395949)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/200/4733c27...beaufortfrancois:7395949.html" title="Last updated on Jun 8, 2026, 4:56 PM UTC (7395949)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -264,8 +264,10 @@ To <dfn for="model context">unregister a tool</dfn> given a {{ModelContext}} |mo
 
 1. [=map/Remove=] |tool map|[|tool name|].
 
-1. Run [=notify documents of a tool change=] given |modelContext|'s [=relevant global object=]'s
-   [=associated Document|associated <code>Document</code>=] and |exposed origins|.
+1. Run the following steps [=in parallel=]:
+
+   1. [=Notify documents of a tool change=] given |modelContext|'s [=relevant global object=]'s
+      [=associated Document|associated <code>Document</code>=] and |exposed origins|.
 
 </div>
 
@@ -448,9 +450,9 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
 
 1. Let |promise| be [=a new promise=] created in [=this=]'s [=relevant realm=].
 
-1. Run the following substeps [=in parallel=]:
+1. Run the following steps [=in parallel=]:
 
-  1. Run [=notify documents of a tool change=] given |tool owner| and |exposed origins|.
+  1. [=Notify documents of a tool change=] given |tool owner| and |exposed origins|.
 
   1. [=Queue a global task=] on the [=webmcp task source=] given |global| to
      [=resolve=] |promise| with undefined.
--- a/index.bs
+++ b/index.bs
@@ -331,8 +331,9 @@ is a [=model context=] [=struct=] created alongside the {{ModelContext}}.
 <div algorithm>
 The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var>)</dfn> method steps are:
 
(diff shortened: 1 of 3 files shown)


index.bs
```diff
@@ -440,7 +442,15 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
 1. Set [=this=]'s [=ModelContext/internal context=][|tool name|] to |tool definition|.
 
-1. Run [=notify documents of a tool change=] given |tool owner| and |exposed origins|.
+1. Let |promise| be [=a new promise=].
+
+1. Run the following substeps in parallel:
+
+  1. Run [=notify documents of a tool change=] given |tool owner| and |exposed origins|.
+
+  1. [=Resolve=] |promise| with undefined.
```
domfarolino: We can't resolve promises in parallel. You have to queue a task back on the global that owns the promise, to resolve it there. See https://html.spec.whatwg.org/multipage/infrastructure.html#parallelism:queue-a-global-task-2 for a good example of this, except we want to use the https://webmachinelearning.github.io/webmcp/#webmcp-task-source instead of the DOM manipulation one.

beaufortfrancois: 🤦 Of course.. I forgot. I'll update this!

beaufortfrancois: Done.

domfarolino:
I pushed some small fixes up:
 - Preserving `then`, per https://github.com/whatwg/html?tab=contributing-ov-file#common-mistakes-around-prose-style which I try and follow in all my specs
 - Fixed the parallelism issue. Before https://github.com/webmachinelearning/webmcp/pull/200/changes/4d6c78bcd9a87cf847d326e69be73a59f5c12e95, this PR broke the assert in https://webmachinelearning.github.io/webmcp/#notify-documents-of-a-tool-change, and that commit fixes it by adjusting the assert. This is right, since we need to queue a task to resolve the promise from an _in parallel_ context, in order to make the Promise resolution well-ordered with the `toolchange` events.
 - Runs the unregistration steps from _in parallel_ too, which is now required since they assert/expect that.

Can you take a look @beaufortfrancois? If it looks good to you, I think we can merge.

bwalderman (approved):

beaufortfrancois:
@domfarolino Thanks for the fixes!

PR #202, opened 2026-06-09
Reject registerTool() promise when signal is aborted

beaufortfrancois:
I believe we should reject `registerTool()` promise when signal is aborted instead now following https://github.com/webmachinelearning/webmcp/pull/200/ changes

It's part of changes required for webmachinelearning/webmcp#175.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/webmcp/pull/202.html" title="Last updated on Jun 17, 2026, 8:30 AM UTC (7526883)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/202/7addecf...beaufortfrancois:7526883.html" title="Last updated on Jun 17, 2026, 8:30 AM UTC (7526883)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -399,9 +399,8 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
 
 1. If |signal| [=map/exists=], then:
 
-   1. If |signal| is [=AbortSignal/aborted=], then optionally [=report a warning to the console=]
-      indicating that the tool was not registered because the {{AbortSignal}} was already
-      [=AbortSignal/aborted=], and return.
+   1. If |signal| is [=AbortSignal/aborted=], then return [=a promise rejected with=]
+      |signal|'s [=AbortSignal/abort reason=].
 
    1. [=AbortSignal/add|Add an abort algorithm=] to |signal| that [=model context/unregisters a
       tool=] given [=this=] and |tool name|.
--- a/index.bs
+++ b/index.bs
@@ -395,6 +395,8 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
 1. Let |untrusted content hint| be true if |tool|'s {{ModelContextTool/annotations}} [=map/exists=] and
    its {{ToolAnnotations/untrustedContentHint}} is true. Otherwise, let it be false.
 
+1. Let |promise| be [=a new promise=] created in [=this=]'s [=relevant realm=].
+
 1. Let |signal| be |options|'s {{ModelContextRegisterToolOptions/signal}}.
 
 1. If |signal| [=map/exists=], then:
@@ -402,8 +404,11 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
    1. If |signal| is [=AbortSignal/aborted=], then return [=a promise rejected with=]
       |signal|'s [=AbortSignal/abort reason=].
 
-   1. [=AbortSignal/add|Add an abort algorithm=] to |signal| that [=model context/unregisters a
-      tool=] given [=this=] and |tool name|.
+   1. [=AbortSignal/add|Add the following abort steps=] to |signal|:
+
+        1. [=model context/Unregister a tool=] given [=this=] and |tool name|.
+
+        1. [=Reject=] |promise| with |signal|'s [=AbortSignal/abort reason=].
 
 1. Let |exposed origins| be an empty [=list=] of [=origins=].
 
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
@domfarolino Please review

index.bs
```diff
@@ -399,9 +399,8 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
 
 1. If |signal| [=map/exists=], then:
 
-   1. If |signal| is [=AbortSignal/aborted=], then optionally [=report a warning to the console=]
-      indicating that the tool was not registered because the {{AbortSignal}} was already
-      [=AbortSignal/aborted=], and return.
+   1. If |signal| is [=AbortSignal/aborted=], then return [=a promise rejected with=]
+      |signal|'s [=AbortSignal/abort reason=].
 
    1. [=AbortSignal/add|Add an abort algorithm=] to |signal| that [=model context/unregisters a
```
beaufortfrancois: I also need to reject promise when tool gets unregistered but it's not clear to me how to do this properly. Should  I add |promise| and |signal| parameters to the "unregister a tool" algorithm or should I call both reject promise and then unregister tool algorithm?

beaufortfrancois: I gave it a shot. Please let me know if that looks good to you.

beaufortfrancois: One thing I'm still not sure is whether we want "Notify documents of a tool change" to run before rejecting the promise when signal is aborted.

beaufortfrancois: @domfarolino The current state of this PR reflects what is implemented in https://chromium-review.googlesource.com/c/chromium/src/+/7916507. Let me know if that looks good to you.

index.bs
```diff
@@ -395,16 +395,20 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
-      indicating that the tool was not registered because the {{AbortSignal}} was already
-      [=AbortSignal/aborted=], and return.
+   1. If |signal| is [=AbortSignal/aborted=], then return [=a promise rejected with=]
+      |signal|'s [=AbortSignal/abort reason=].
+
+   1. [=AbortSignal/add|Add the following abort steps=] to |signal|:
 
-   1. [=AbortSignal/add|Add an abort algorithm=] to |signal| that [=model context/unregisters a
-      tool=] given [=this=] and |tool name|.
+        1. [=model context/Unregister a tool=] given [=this=] and |tool name|.
```
domfarolino, line 409: I think there's a bug here. If we set up this Promise with the abort steps to call https://webmachinelearning.github.io/webmcp/#model-context-unregister-a-tool, but then we return a rejected Promise because we fail the `|exposed origins|` check, but later the author rejects the abort signal, then we still end up running https://webmachinelearning.github.io/webmcp/#model-context-unregister-a-tool which fails the assert that the tool exists in the registry, right?

Maybe we should change https://webmachinelearning.github.io/webmcp/#model-context-unregister-a-tool to match the Chromium implementation, which first [checks if the doomed tool exists](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/script_tools/model_context.cc;l=388-391;drc=71d8f4a1071b2e13b9985b249ba1ec8e4bde8ae6). How does that sound?

index.bs
domfarolino: > One thing I'm still not sure is whether we want "Notify documents of a tool change" to run before rejecting the promise when signal is aborted.

Can you clarify the scenario? If we actually end up unregistering a tool, that means the the tool was successfully added to the tool map, and the Promise should already be resolved and can't be rejected, right? I guess not, because we could still be in parallel waiting for the Promise resolution task to be queued, and we could _beat it_ by first aborting the signal on the main thread immediately after `registerTool()`. I think what you have right now is good: we call unregister, which _asynchronously_ sends the "notify documents of a tool chance" IPC / task, and then _synchronously_ rejects the Promise. That's good because it makes it impossible for a tool to be unregistered correctly but have the Promise be resolved instead of rejected.

So I think what we have here is "right" and indeed matches the impl as you said. Do the details I've spewed above match your mental model?

index.bs
beaufortfrancois: That's a good catch. Thank you!

index.bs
beaufortfrancois: It makes sense to me as well. Asynchronicity can be hard to spec sometimes and I wanted your review on this.

NoClueMike (approved):

domfarolino (approved):

Issue #165, opened 2026-04-16
WebMCP elicitation via requestUserInteraction() (was: Human in the Loop support for non-browser clients)


MiguelsPizza:
The current spec is geared toward [Human-in-the-Loop](https://webmachinelearning.github.io/webmcp/#dom-modelcontextclient-requestuserinteraction) which is unofficially interpreted as "the agent lives in the context of the browser" (browser's agent, extension agent, Iframe agent etc). We initially elected that external agent browser automation cases be out of scope (see #21  and the [2025-09-18 resolution](https://www.w3.org/2025/10/02-webmachinelearning-minutes.html)) and that WebMCP focuses on human-in-the-loop use cases initially. However, from what I've seen, a large part of the excitement around this API is geared toward remote browser automation over things like WebDriver BiDi / CDP.

Source: [Cloudflare supports WebMCP-enabled browsers on their Browser Run platform](https://developers.cloudflare.com/browser-run/features/webmcp/)

<img width="800" height="300" alt="Image" src="https://github.com/user-attachments/assets/7617ec3c-d8ec-462f-9f32-0302e205d585" />

They also released an interesting [Human-in-the-Loop feature](https://developers.cloudflare.com/browser-run/features/human-in-the-loop/) that got me thinking about what a first-party integration between WebMCP and a non-browser-native client might look like. The user might be present but driving from outside the browser, so does this fall into our Human-in-the-Loop domain?

It might look something like this: 

<img width="800" height="500" alt="Image" src="https://github.com/user-attachments/assets/8405dc98-9843-49b3-a6bf-38c11ccb31f1" />

Idan-Levin:
Hi @MiguelsPizza,

I think there's an important question in the business domain to work through before the remote browser implementation, which is: do website owners actually want their site to be executable remotely?

I believe in many cases the answer would be no. E-commerce is a good example, a storefront owner would love users running an AI side panel while they're browsing his site,  but I'm not sure he'd like the ability to scrape the catalog or perform operations on the site without human eyes present. And this eventually might prevent many of them to implement WebMCP. 

So the question is can there be different treatment for a remote browser vs. a browser where a human is present?

MiguelsPizza:
I appreciate your thoughts @Idan-Levin 

>  do website owners actually want their site to be executable remotely

[...]

PR #204, opened 2026-06-09
Replace requestUserInteraction with requestUserInput

bwalderman:
Update the spec to reflect the latest design proposal for elicitation described in #165  

This replaces `requestUserInteraction` on `ModelContextClient` with a new `requestUserInput` method which supports three elicitation modes:
- interactive: In-page UI flow for local browsing (replaces the previous proposed requestUserInteraction behavior)
- form: Structured data collection via JSON Schema, works for both local and remote browsing scenarios
- url: Out-of-band URL navigation for sensitive flows (auth, payment), works for both local and remote browsing scenarios

The form and URL modes are aligned with MCP Elicitation (https://modelcontextprotocol.io/specification/draft/client/elicitation).

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/204.html" title="Last updated on Jun 9, 2026, 7:06 PM UTC (27d4013)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/204/b8617aa...27d4013.html" title="Last updated on Jun 9, 2026, 7:06 PM UTC (27d4013)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -561,28 +561,105 @@ dictionary ModelContextRegisterToolOptions {
 
 <h4 id="model-context-client">ModelContextClient Interface</h4>
 
-The {{ModelContextClient}} interface represents an [=agent=] executing a tool provided by the site through the {{ModelContext}} API.
+The {{ModelContextClient}} interface represents an [=agent=] executing a tool provided by the site
+through the {{ModelContext}} API.
 
 <xmp class="idl">
 [Exposed=Window, SecureContext]
 interface ModelContextClient {
-  Promise<any> requestUserInteraction(UserInteractionCallback callback);
+  Promise<ModelContextUserInputResult> requestUserInput(
+      ModelContextUserInputOptions options,
+      optional ModelContextInteractiveCallback callback);
 };
 
-callback UserInteractionCallback = Promise<any> ();
+callback ModelContextInteractiveCallback = Promise<any> ();
+
+enum ModelContextUserInputMode { "interactive", "form", "url" };
+
+enum ModelContextUserInputAction { "accept", "decline", "cancel" };
+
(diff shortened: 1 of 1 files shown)


bwalderman:
Algorithm steps are still TODO. I'll fill these in with a follow-up PR. This PR is just to get the interfaces in.

bwalderman:
CI / Build failure appears to be a bug in the latest `bikeshed`. Opened a PR to fix: https://github.com/speced/bikeshed/pull/3309

domfarolino:
We're planning on discussing https://github.com/webmachinelearning/webmcp/issues/165 more in tomorrow's call anyways, but I will say that from talking about this a bit internally, we're more interested in supporting the `mode: 'interactive'` version for a number of reasons, compared to the others, so I might prefer we pursue only that approach initially.

> Algorithm steps are still TODO. I'll fill these in with a follow-up PR. This PR is just to get the interfaces in.

Also I think we should avoid filling the spec with incomplete interfaces and empty algorithms. There can be subordinate algorithms to a larger feature that have big TODOs in them in a draft spec of course, but in general I think we should stray away from adding stubs to the spec and filling them out later. This is the same reason that `getTools()` and `executeTool()` don't exist as empty methods yet before their actual meat gets filled out—makes sense to just do them all at once.

beaufortfrancois:
Shall we remove the [ModelContextClient Interface](https://webmachinelearning.github.io/webmcp/#model-context-client) from the spec for now then until we've identified the proper replacement?

domfarolino:
Yeah I think that's reasonable: https://github.com/webmachinelearning/webmcp/pull/205

PR #217, opened 2026-07-01
Add consequentialHint to ToolAnnotations

### LINKED ISSUE #176: Hint for reversible or consequential actions

beaufortfrancois:
Fix https://github.com/webmachinelearning/webmcp/issues/176

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/webmcp/pull/217.html" title="Last updated on Jul 24, 2026, 12:53 PM UTC (20de666)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/217/5801678...beaufortfrancois:20de666.html" title="Last updated on Jul 24, 2026, 12:53 PM UTC (20de666)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -177,6 +177,9 @@ A <dfn>tool definition</dfn> is a [=struct=] with the following [=struct/items=]
   : <dfn>untrusted content hint</dfn>
   :: a [=boolean=], initially false.
 
+  : <dfn>consequential hint</dfn>
+  :: a [=boolean=], initially false.
+
   : <dfn>exposed origins</dfn>
   :: a [=list=] or [=origins=], initially [=list/empty=].
 </dl>
@@ -395,6 +398,9 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
 1. Let |untrusted content hint| be true if |tool|'s {{ModelContextTool/annotations}} [=map/exists=] and
    its {{ToolAnnotations/untrustedContentHint}} is true. Otherwise, let it be false.
 
+1. Let |consequential hint| be true if |tool|'s {{ModelContextTool/annotations}} [=map/exists=] and
+   its {{ToolAnnotations/consequentialHint}} is true. Otherwise, let it be false.
+
 1. Let |promise| be [=a new promise=] created in [=this=]'s [=relevant realm=].
 
 1. Let |signal| be |options|'s {{ModelContextRegisterToolOptions/signal}}.
@@ -447,6 +453,9 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
    : [=tool definition/untrusted content hint=]
    :: |untrusted content hint|
 
+   : [=tool definition/consequential hint=]
+   :: |consequential hint|
+
    : [=tool definition/exposed origins=]
    :: |exposed origins|
 
@@ -482,6 +491,7 @@ dictionary ModelContextTool {
 dictionary ToolAnnotations {
   boolean readOnlyHint = false;
   boolean untrustedContentHint = false;
+  boolean consequentialHint = false;
 };
 
 callback ToolExecuteCallback = Promise<any> (object input);
@@ -531,6 +541,9 @@ The {{ToolAnnotations}} dictionary provides optional metadata about a tool:
 
   : <code><var ignore>annotations</var>["{{ToolAnnotations/untrustedContentHint}}"]</code>
   :: If true, indicates that the tool's output contains data that is untrusted, from the perspective of the author registering the tool.
+
+  : <code><var ignore>annotations</var>["{{ToolAnnotations/consequentialHint}}"]</code>
+  :: If true, indicates that executing the tool will result in consequential actions, ex: booking a flight, transferring money.
 </dl>
 
 
@@ -1100,6 +1113,14 @@ respective private browsing modes are safely exposed to [=agents=] and that thes
 **Threats addressed:** [[#prompt-injection]] ([[#output-injection-attacks]])
 
 **How:** A boolean {{ToolAnnotations/untrustedContentHint}} annotation that acts as a signal to the client that the payload requires heightened security handling, allowing the client to sanitize the payload, use indicators such as spotlighting [[SPOTLIGHTING]] to highlight untrustworthy content to the model, or hide that part of the response entirely.
+
+<h4 id="mitigation-consequential-annotation">Consequential Annotation for Tool Executions</h4>
+
+**What:** Providing agents with a signal that a tool's execution results in significant, real-world, or non-reversible consequences.
+
+**Threats addressed:** [[#misrepresentation-of-intent]]
+
+**How:** A boolean {{ToolAnnotations/consequentialHint}} annotation acts as a signal to the client or agent that the tool performs a consequential action, such as booking a flight or transferring money. This way they can selectively enforce mandatory user confirmation prompts before executing high-stakes tools, directly mitigating the risk of accidental or malicious misrepresentation of intent.
 
 <h2 id="accessibility">Accessibility considerations</h2>
 
--- a/index.bs
+++ b/index.bs
@@ -669,7 +669,7 @@ The {{ToolAnnotations}} dictionary provides optional metadata about a tool:
   :: If true, indicates that the tool's output contains data that is untrusted, from the perspective of the author registering the tool.
 
   : <code><var ignore>annotations</var>["{{ToolAnnotations/consequentialHint}}"]</code>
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
@domfarolino Please review

NoClueMike (commented):
When did "hints" deserve the time spent over actual communication. In the clients world, they can't even enable api"s for tools.

index.bs
```diff
@@ -1101,6 +1114,14 @@ respective private browsing modes are safely exposed to [=agents=] and that thes
 
 **How:** A boolean {{ToolAnnotations/untrustedContentHint}} annotation that acts as a signal to the client that the payload requires heightened security handling, allowing the client to sanitize the payload, use indicators such as spotlighting [[SPOTLIGHTING]] to highlight untrustworthy content to the model, or hide that part of the response entirely.
 
+<h4 id="mitigation-consequential-annotation">Consequential Annotation for Tool Executions</h4>
+
+**What:** Providing agents with a signal that a tool's execution results in significant, real-world, or non-reversible consequences.
+
+**Threats addressed:** [[#misrepresentation-of-intent]]
+
+**How:** A boolean {{ToolAnnotations/consequentialHint}} annotation acts as a signal to the client or agent that the tool performs a consequential action, such as booking a flight or transferring money. This way they can selectively enforce mandatory user confirmation prompts before executing high-stakes tools, directly mitigating the risk of accidental or malicious misrepresentation of intent.
```
victorhuangwq, line 1320: I agree that consequentialHint helps with the accidental case. Using the finalizeCart example, the name and description may be ambiguous about whether the tool merely prepares the cart or completes the purchase. A non-malicious developer still knows that execute triggers a purchase, so they will likely be marking the tool as consequential. This would give the client or agent an explicit signal that it should request confirmation rather than relying only on its interpretation of “finalize.”

I’m less convinced it mitigates the malicious case described here. A malicious developer controls the name, description, implementation, and annotations. If they deliberately describe finalizeCart ambiguously to conceal a purchase, they will omit consequentialHint or set it to false.

victorhuangwq (commented):
@beaufortfrancois thanks for adding this!

On a broader note, I'm starting to think if all ToolAnnotations should default to `null` instead of just false? 

For example, I'm not sure if `consequential` not being set is the same as `false`.

index.bs
johannhof: Yeah it does not prevent "malicious misrepresentation of intent", but beyond preventing mistakes it can help defend against indirect prompt injection that wants to trigger these actions, since the user would be asked to confirm them.

Maybe we should say that instead :)

johannhof:
Is `null` an option? I agree it seems better.

domfarolino:
If we don't like the idea of defaulting hints to `false`, we could provide no default at all; in that case they'd just be missing from the ToolAnnotations dictionary (`undefined`, just like `annotations.fakeHint`) and wouldn't be provided to the model (or the model would be expecting them, and would have to "fill in" _false_).

But then we'd have three values to think about: `undefined`, `true`, and `false`. Tool authors might ask when to explicitly pass `false` vs. leave it `undefined`, and tool consumers (models) now need to think about whether there's significance in the third value—did the developer just mean to omit the hint, but passed in `false` explicitly? Or does _explicit `false`_ mean something special that `undefined` doesn't capture??

The required/default boolean sidesteps this noise, so I like it more. I think it just matters how this will be used by models, but I think the binary makes more sense and models are unlikely to make use of all three values.

index.bs
domfarolino: +1 I agree with that framing, and we should avoid over-promising on malicious intent.

beaufortfrancois: I've updated this section with your feedback. Thanks!

victorhuangwq:
My original concern was that from an agent developer's perspective, it's an easy mistake to read consequentialHint = false as "this is safe," when it might just mean the web developer forgot to set it (or didn't know they should). A third "unset" state would mitigate that, though I take your point about the noise a special third value introduces.

I agree that making it required sidesteps the "forgot to set" problem entirely. If we go that route, do we intend it to be required for all ToolAnnotations, or only a certain subset? (Btw I don't think this discussion should block the current PR)

index.bs
beaufortfrancois: And I just rebased it.

index.bs
```diff
@@ -1299,6 +1311,14 @@ respective private browsing modes are safely exposed to [=agents=] and that thes
 
 **How:** A boolean {{ToolAnnotations/untrustedContentHint}} annotation that acts as a signal to the client that the payload requires heightened security handling, allowing the client to sanitize the payload, use indicators such as spotlighting [[SPOTLIGHTING]] to highlight untrustworthy content to the model, or hide that part of the response entirely.
 
+<h4 id="mitigation-consequential-annotation">Consequential Annotation for Tool Executions</h4>
+
+**What:** Providing agents with a signal that a tool's execution results in significant, real-world, or non-reversible consequences.
```
bvandersloot-mozilla, line 1316: There are more cases covered here than on line 670 where the dictionary fields are defined. These should align. Probably with this, more detailed, one.

bvandersloot-mozilla:
A ternary on the flag seems unwise. Particularly when undefined is falsy.

Assigning the safe default would be better than making them required, and is probably better than the current unsafe default (untrustedContentHint = false and consequentialHint = false) . Particularly because the failure cases have pretty different visibility and severity, e.g. increased friction vs silently taking a consequential action.

index.bs
beaufortfrancois: It's not clear to me what you suggest.

index.bs
```diff
@@ -658,6 +667,9 @@ The {{ToolAnnotations}} dictionary provides optional metadata about a tool:
 
   : <code><var ignore>annotations</var>["{{ToolAnnotations/untrustedContentHint}}"]</code>
   :: If true, indicates that the tool's output contains data that is untrusted, from the perspective of the author registering the tool.
+
+  : <code><var ignore>annotations</var>["{{ToolAnnotations/consequentialHint}}"]</code>
+  :: If true, indicates that executing the tool will result in consequential actions, ex: booking a flight, transferring money.
```
bvandersloot-mozilla:
```suggestion
  :: If true, indicates that executing the tool will result in consequential actions that are significant, real-world, or non-reversible, ex: booking a flight, transferring money.
```

index.bs
bvandersloot-mozilla: Suggestion inline above: https://github.com/webmachinelearning/webmcp/pull/217/changes#r3645416367

beaufortfrancois: Thanks!

victorhuangwq:
@bvandersloot-mozilla if`consequentialHint` is kept optional, we should still default to `false`, as per the discussion in #176 , since if it is on by default it will likely dilute what the signal tells the receiver about actual intent.

Like @domfarolino suggested in https://github.com/webmachinelearning/webmcp/issues/176#issuecomment-4507743028, most tools are likely going to be inconsequential. So having `consequentialHint=true` will mean the developer took deliberate action to indicate that.

`undefined` being falsy is a fair point for the ternary case, though my thought was that careful agent implementers at least will at least have a way to tell "not set" apart from "deliberately false".

I think it's better to revisit once we have Origin Trial feedback.

johannhof:
Yeah I think I'm coming around to the argument that ternary isn't desirable. The agent will have to make a binary decision either way with the information it is given. But I agree with Victor here that consequential = true by default seems to risk polluting / diluting the signal.

beaufortfrancois:
Since `consequentialHint` has landed in Chromium https://chromiumdash.appspot.com/commit/a1bf6d8347ecd82fae080076f62a61e8795d8e22, shall we merge this PR or wait?

domfarolino (approved):

PR #223, opened 2026-07-17
Add `getTools()`

domfarolino:
This PR specifies the `getTools()` API. As such, this makes progress on all of the following:

 - https://github.com/webmachinelearning/webmcp/issues/57
 - https://github.com/webmachinelearning/webmcp/issues/117
 - https://github.com/webmachinelearning/webmcp/issues/159
 - https://github.com/webmachinelearning/webmcp/issues/160

All of these issues can be closed when the next PR that specified `executeTool()` lands.

/cc @beaufortfrancois who will be interested in reviewing, but who I cannot formally request a review from (we'll fix that).

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/223.html" title="Last updated on Jul 21, 2026, 6:03 PM UTC (a603abe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/223/d29b8d7...a603abe.html" title="Last updated on Jul 21, 2026, 6:03 PM UTC (a603abe)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -721,26 +721,35 @@ dictionary RegisteredTool {
 
 <dl class="domintro">
   : <code><var ignore>tool</var>["{{RegisteredTool/name}}"]</code>
-  :: A unique identifier for the tool. It is the same value provided at tool registration, via {{ModelContextTool/name}}.
+  :: A unique identifier for the tool. It is the same value provided at tool registration, via
+     {{ModelContextTool/name}}.
 
   : <code><var ignore>tool</var>["{{RegisteredTool/title}}"]</code>
-  :: A human-readable label for the tool. It is the same value provided at tool registration, via {{ModelContextTool/title}}.
+  :: A human-readable label for the tool. It is the same value provided at tool registration, via
+     {{ModelContextTool/title}}.
 
   : <code><var ignore>tool</var>["{{RegisteredTool/description}}"]</code>
-  :: A natural language description of the tool's functionality. It is the same value provided at tool registration, via {{ModelContextTool/description}}.
+  :: A natural language description of the tool's functionality. It is the same value provided at
+     tool registration, via {{ModelContextTool/description}}.
 
   : <code><var ignore>tool</var>["{{RegisteredTool/inputSchema}}"]</code>
   :: A stringified JSON Schema object describing the expected input parameters for the tool
-     [[!JSON-SCHEMA]]. It is the string derived from {{ModelContextTool/inputSchema}} during registration.
+     [[!JSON-SCHEMA]]. It is the string derived from {{ModelContextTool/inputSchema}} during
+     registration.
 
   : <code><var ignore>tool</var>["{{RegisteredTool/window}}"]</code>
   :: The {{Window}} of the document that registered the tool.
 
   : <code><var ignore>tool</var>["{{RegisteredTool/origin}}"]</code>
-  :: The origin of the document that registered the tool. This member is only meaningful when the tool is cross-origin, and the consumer of a tool cannot otherwise get the tool's origin from its {{RegisteredTool/window}}. For same-origin tools, this is the same as the tool's {{RegisteredTool/window}}'s {{WindowOrWorkerGlobalScope/origin}}, and the caller's own {{Window}}.{{WindowOrWorkerGlobalScope/origin}}.
+  :: The origin of the document that registered the tool. This member is only meaningful when the
+     tool is cross-origin, and the consumer of a tool cannot otherwise get the tool's origin from
+     its {{RegisteredTool/window}}. For same-origin tools, this is the same as the tool's
+     {{RegisteredTool/window}}'s {{WindowOrWorkerGlobalScope/origin}}, and the caller's own
+     {{Window}}.{{WindowOrWorkerGlobalScope/origin}}.
 
   : <code><var ignore>tool</var>["{{RegisteredTool/annotations}}"]</code>
-  :: Optional annotations providing metadata about the tool. It matches {{ModelContextTool/annotations}}.
+  :: Optional annotations providing metadata about the tool. It matches
+     {{ModelContextTool/annotations}}.
 </dl>
 
 
--- a/index.bs
+++ b/index.bs
@@ -710,7 +710,7 @@ dictionary RegisteredTool {
   // `title` can be exposed as a `DOMString` since it was taken in by a
   // `USVString`, meaning all unmatched surrogate processing has already been
   // done, and there's no need to do it again on tool exposure.
-  required DOMString title;
+  DOMString title;
   required DOMString description;
   DOMString inputSchema;
   required Window window;
--- a/index.bs
+++ b/index.bs
@@ -515,8 +515,8 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
 
    1. Let |tools| be an empty [=list=] of {{RegisteredTool}} dictionaries.
 
-   1. Let |navigables| be |toolRequestor|'s [=node navigable=]'s [=navigable/traversable navigable=]'s
-      [=Document/inclusive descendant navigables=].
+   1. Let |navigables| be |toolRequestor|'s [=node navigable=]'s [=navigable/traversable
+      navigable=]'s [=Document/inclusive descendant navigables=].
 
    1. [=list/For each=] |navigable| of |navigables|:
 
@@ -529,11 +529,10 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
 
       1. Let |callerOrigin| be |toolRequestor|'s [=Document/origin=].
 
-      1. Let |is visible to caller| be true if |targetOrigin| is [=same origin=] with
-         |callerOrigin|, or if |from origins| [=list/contains=] |targetOrigin|. Otherwise, let it be
-         false.
+      1. If |toolOwnerIsRequested| be true if |targetOrigin| is [=same origin=] with |callerOrigin|,
+         or if |from origins| [=list/contains=] |targetOrigin|. Otherwise, let it be false.
 
-      1. If |is visible to caller| is false, then [=iteration/continue=].
+      1. If |toolOwnerIsRequested| is false, then [=iteration/continue=].
 
       1. Let |targetToolMap| be |targetDocument|'s [=Document/associated ModelContext|associated
          <code>ModelContext</code>=]'s [=ModelContext/internal context=]'s [=model context/tool
--- a/index.bs
+++ b/index.bs
@@ -572,10 +572,11 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
             :: |targetOrigin|, [=serialization of an origin|serialized=].
 
             : {{RegisteredTool/annotations}}
-            :: A new {{ToolAnnotations}} dictionary whose {{ToolAnnotations/readOnlyHint}} is |tool
-               definition|'s [=tool definition/read-only hint=] and
+            :: if |tool definition|'s [=tool definition/annotations=] is not null, a
+               {{ToolAnnotations}} dictionary whose {{ToolAnnotations/readOnlyHint}} is |tool
+               definition|'s [=tool definition/annotations=]'s [=annotations/read-only hint=] and
                {{ToolAnnotations/untrustedContentHint}} is |tool definition|'s [=tool
-               definition/untrusted content hint=].
+               definition/annotations=]'s [=annotations/untrusted content hint=].
 
          1. [=list/Append=] |registeredTool| to |tools|.
 
--- a/index.bs
+++ b/index.bs
@@ -549,7 +549,10 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
             :: |tool definition|'s [=tool definition/name=]
 
             : {{RegisteredTool/title}}
-            :: |tool definition|'s [=tool definition/title=]
+            :: |tool definition|'s [=tool definition/title=] if it is non-null; otherwise the empty string.
+
+               Issue(#224): Consider not defaulting to the empty string, and just excluding this
+               member, which will result in {{undefined}}.
 
             : {{RegisteredTool/description}}
             :: |tool definition|'s [=tool definition/description=]
@@ -706,7 +709,7 @@ dictionary RegisteredTool {
   // `title` can be exposed as a `DOMString` since it was taken in by a
   // `USVString`, meaning all unmatched surrogate processing has already been
   // done, and there's no need to do it again on tool exposure.
-  DOMString title;
+  required DOMString title;
   required Window window;
   required USVString origin;
   ToolAnnotations annotations;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -464,6 +471,111 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
+            :: |tool definition|'s [=tool definition/name=]
+
+            : {{RegisteredTool/title}}
+            :: |tool definition|'s [=tool definition/title=]
+
+            : {{RegisteredTool/description}}
+            :: |tool definition|'s [=tool definition/description=]
+
+            : {{RegisteredTool/inputSchema}}
+            :: |tool definition|'s [=tool definition/input schema=] if it is not the empty string;
```
beaufortfrancois, line 565: It'd be nice to be consistent between `inputSchema` and `title` since both are strings. We should either have them return `undefined` if it is the empty string, or return `''`. Right now, in Chromium, a tool registered without `inputSchema` not `title` has `getTools` returns an array with a tool dictionary that contains an empty string `title` key but no `inputSchema` key.

index.bs
```diff
@@ -464,6 +471,111 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
+      1. Let |targetDocument| be |navigable|'s [=navigable/active document=].
+
+      1. If |targetDocument| is not [=allowed to use=] the "{{tools}}" feature, then
+         [=iteration/continue=].
+
+      1. Let |targetOrigin| be |targetDocument|'s [=Document/origin=].
+
+      1. Let |callerOrigin| be |toolRequestor|'s [=Document/origin=].
+
+      1. If |toolOwnerIsRequested| be true if |targetOrigin| is [=same origin=] with |callerOrigin|,
```
beaufortfrancois, line 534: Did you mean `Let |toolOwnerIsRequested| be true` instead of `If |toolOwnerIsRequested| be true`?

index.bs
```diff
@@ -464,6 +471,111 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
+         or if |from origins| [=list/contains=] |targetOrigin|. Otherwise, let it be false.
+
+      1. If |toolOwnerIsRequested| is false, then [=iteration/continue=].
```
beaufortfrancois, line 537: We may want to simplify things and say something like "If targetOrigin is not same origin with callerOrigin, and fromOrigins does not contain targetOrigin, then continue."

index.bs
```diff
@@ -464,6 +471,111 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
+
+      1. Let |targetToolMap| be |targetDocument|'s [=Document/associated ModelContext|associated
+         <code>ModelContext</code>=]'s [=ModelContext/internal context=]'s [=model context/tool
+         map=].
+
+      1. [=map/For each=] <var ignore>tool name</var> → |tool definition| of |targetToolMap|:
+
+         1. If [=tool is exposed to an origin=] given |targetOrigin|, |tool definition|'s [=tool
+            definition/exposed origins=], and |callerOrigin| is false, then [=iteration/continue=].
```
beaufortfrancois: I had trouble reading `and |callerOrigin| is false` first. Can we say something like `returns false` instead of `is false`?

index.bs
```diff
@@ -464,6 +471,111 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
+               otherwise undefined.
+
+            : {{RegisteredTool/window}}
+            :: |targetDocument|'s [=relevant global object=]
+
+            : {{RegisteredTool/origin}}
+            :: |targetOrigin|, [=serialization of an origin|serialized=].
+
+            : {{RegisteredTool/annotations}}
+            :: A new {{ToolAnnotations}} dictionary whose {{ToolAnnotations/readOnlyHint}} is |tool
```
beaufortfrancois: That's not what we have in Chromium, we only set it if tool annotations was set during registration.

index.bs
```diff
@@ -508,7 +620,8 @@ callback ToolExecuteCallback = Promise<any> (object input);
 
   <dt><code><var ignore>tool</var>["{{ModelContextTool/inputSchema}}"]</code></dt>
   <dd>
-    <p>A JSON Schema [[!JSON-SCHEMA]] object describing the expected input parameters for the tool.
+    <p>A stringified JSON Schema object describing the expected input parameters for the tool
```
beaufortfrancois: I believe it is an object, not a stringified version there.

index.bs
```diff
@@ -556,6 +669,68 @@ dictionary ModelContextRegisterToolOptions {
...
 </dl>
 
+<h4 id="model-context-get-tool-options">ModelContextGetToolOptions Dictionary</h4>
+
+The {{ModelContextGetToolOptions}} dictionary allows web applications to filter the tools returned
+by {{ModelContext/getTools()}}.
+
+<xmp class="idl">
+dictionary ModelContextGetToolOptions {
+  sequence<USVString> fromOrigins;
```
beaufortfrancois, line 692: Nit: A JS example of how to use it with iframes for instance would be nice in the spec.

index.bs
```diff
@@ -556,6 +669,68 @@ dictionary ModelContextRegisterToolOptions {
...
+};
+</xmp>
+
+<dl class="domintro">
+  : <code><var ignore>options</var>["{{ModelContextGetToolOptions/fromOrigins}}"]</code>
+  :: An array of origins from which to list tools. When this array is empty or omitted, only tools
```
beaufortfrancois: I'd also add that when it is provided and non-empty, tools in documents whose origins match at least one in the list are included. In addition, same-origin tools are also included.

index.bs
```diff
@@ -556,6 +669,68 @@ dictionary ModelContextRegisterToolOptions {
...
+     [[!JSON-SCHEMA]].
+
+  : <code><var ignore>tool</var>["{{RegisteredTool/title}}"]</code>
+  :: A human-readable label for the tool.
+
+  : <code><var ignore>tool</var>["{{RegisteredTool/window}}"]</code>
+  :: The {{Window}} of the document that registered the tool.
+
+  : <code><var ignore>tool</var>["{{RegisteredTool/origin}}"]</code>
+  :: The origin of the document that registered the tool.
```
beaufortfrancois: Nit: How about explaining that `origin` is exposed only for cross-origin iframes cases otherwise, people could get it from `window.origin`?

index.bs
```diff
@@ -556,6 +669,68 @@ dictionary ModelContextRegisterToolOptions {
...
+  // `USVString`, meaning all unmatched surrogate processing has already been
+  // done, and there's no need to do it again on tool exposure.
+  DOMString title;
+  required Window window;
+  required USVString origin;
+  ToolAnnotations annotations;
+};
+</xmp>
+
+<dl class="domintro">
```
beaufortfrancois, line 723: Shall we link back to https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/223.html#dictdef-modelcontexttool when possible for each key?

index.bs
```diff
@@ -556,6 +669,68 @@ dictionary ModelContextRegisterToolOptions {
...
+  : <code><var ignore>options</var>["{{ModelContextGetToolOptions/fromOrigins}}"]</code>
+  :: An array of origins from which to list tools. When this array is empty or omitted, only tools
+     exposed to documents that are same-origin with the caller of {{ModelContext/getTools()}} are
+     included in the result.
+</dl>
+
+<h4 id="registered-tool">RegisteredTool Dictionary</h4>
+
+The {{RegisteredTool}} dictionary represents a tool that has been registered and is available for
+invocation.
```
beaufortfrancois: Shall we `execution` instead of `invocation` since we'll have `executeTool` soon?

index.bs
```diff
@@ -556,6 +669,68 @@ dictionary ModelContextRegisterToolOptions {
...
+
+<xmp class="idl">
+dictionary RegisteredTool {
```
beaufortfrancois, line 709: Nit: Shall we move all required keys together? Or match order of ModelContextTool dictionary keys?

index.bs
```diff
@@ -464,6 +471,111 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
+
+      1. [=list/Append=] |parsedURL|'s [=url/origin=] to |from origins|.
+
+1. Let |promise| be [=a new promise=] created in [=this=]'s [=relevant realm=].
+
+1. Run the following steps [=in parallel=]:
+
+   1. Let |tools| be an empty [=list=] of {{RegisteredTool}} dictionaries.
+
+   1. Let |navigables| be |toolRequestor|'s [=node navigable=]'s [=navigable/traversable
```
beaufortfrancois, line 520: Is it possible to write this instead?

```
Let navigables be the inclusive descendant navigables of toolRequestor.
```

If so, we should update "notify documents of a tool change" section as well.

index.bs
```diff
@@ -331,6 +332,12 @@ is a [=model context=] [=struct=] created alongside the {{ModelContext}}.
     {{ModelContextTool/description}} are empty strings, or if the {{ModelContextTool/inputSchema}}
     is invalid.</p>
   </dd>
+
+  <dt><code><var ignore>document</var>.{{Document/modelContext}}.{{ModelContext/getTools(options)}}</code></dt>
+  <dd>
+    <p>Returns a promise that resolves to a list of registered tools from this document and its
```
beaufortfrancois, line 346: Shall we precise that getTools is meant for in-page agents, not browser agents which don't need this?

index.bs
```diff
@@ -464,6 +471,111 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
+1. If |toolRequestor| is not [=Document/fully active=], then return [=a promise rejected with=] an
+   "{{InvalidStateError}}" {{DOMException}}.
+
+1. If this's [=surrounding agent=]'s [=agent cluster=]'s [=is origin-keyed=] is false and this's
+   [=relevant settings object=]'s [=environment settings object/origin=]'s [=origin/scheme=] is not
+   <code>"file"</code>, then return [=a promise rejected with=] a "{{SecurityError}}"
+   {{DOMException}}.
+
+1. If |toolRequestor| is not [=allowed to use=] the "{{tools}}" feature, then return [=a promise
+   rejected with=] a "{{NotAllowedError}}" {{DOMException}}.
```
beaufortfrancois, line 498: Shall we move these checks into a dedicated "Is Model Context Allowed" algo that we can reuse for getTools and registerTool? We'll have the same for executeTool soon.

beaufortfrancois (changes requested):

index.bs
domfarolino: Yeah that's reasonable. In https://crrev.com/c/7742044 and https://chat.google.com/room/AAAA92_89dA/teSKVEfp2J4/teSKVEfp2J4?cls=10 (internal only) we decided that it'd be best to make `title` _always_ present, defaulting to empty when it wasn't provided by the developer, for simplicity. At the same time, it doesn't match MCP. I've filed https://github.com/webmachinelearning/webmcp/issues/224 and for now just make sure this is aligned with the implementation.

index.bs
domfarolino: Yep.

index.bs
domfarolino: It shrinks the number of lines, but it does invert both conditions and forces them into having negatives, and negative logical conditions are cognitively harder to parse than positive ones, which is why I framed it slightly more verbose here, but avoided as many negatives as I could. So if it's alright, I'd prefer to keep this!

index.bs
domfarolino: Yeah, good feedback. Done.

index.bs
domfarolino: Hmm, but that would only fetch the tools *under* the requestor's subtree, right? They'd miss out on any tools from Documents "above" them in the frame tree that are explicitly exposed to the requestor's origin.

index.bs
domfarolino: Done!

index.bs
domfarolino: Sure. Done.

index.bs
domfarolino: Good call. I've matched the order of `ModelContextTool`, including leaving `annotations` last.

index.bs
domfarolino: But it's not, is it? I think it's [always exposed](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/script_tools/model_context.cc;l=998;drc=e96168449af0d831581c8e5dbedf5ae171cf6120) in the impl and [the spec](https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/223.html#ref-for-dom-registeredtool-origin). No?

index.bs
domfarolino: Oops, you're right. Done.

index.bs
beaufortfrancois: I meant we're exposing `origin` specifically only because of cross-origin iframes right? Web developers could get it directly from `window.origin` if we were to target only same-origin iframes.

index.bs
domfarolino: I've re-worded this a good bit to describe that this doesn't control which tools make it in, but rather which documents from which to query tools. Lmk what you think.

index.bs
domfarolino: True it's useful only for cross-origin  tools. But it's still exposed via `RegisteredTool` either way. We never omit it.

You wrote "How about explaining that `origin` is exposed only for cross-origin iframes" and I think I interpreted this as "How about explaining that we only ever expose `origin` to JavaScript when the tool is cross-origin", but now I think you meant "How about explaining that the purpose of exposing `origin` is really for cross-origin tools".

If that's right, I can modify the description here to explain the intent.

domfarolino: Done. Lmk if you think the updated description looks alright.

beaufortfrancois: Exactly!

index.bs
domfarolino: Good idea. Done.

index.bs
domfarolino: I think that's reasonable, it's just slightly tedious since the algorithm would I guess either return a "rejected promise-or-null" or a "DOMException name-or-null", and then the caller would have to reject the Promise with the non-null value if it exists, and do nothing otherwise. It adds another layer of indirection which is slightly weird, but what do you think?

index.bs
domfarolino: I've rebased this on top of #225 and I think this matches Chromium now, PTAL!

index.bs
beaufortfrancois: Let's keep it like that for now and revisit if we find a nice pattern.

index.bs
beaufortfrancois: I was wrong. Let's resolve this comment.

index.bs
beaufortfrancois: WDYT?

beaufortfrancois (approved):

PR #225, opened 2026-07-21
Move `annotations` into a separate object, in preparation for `getTools()`

domfarolino:
This is pre-work for https://github.com/webmachinelearning/webmcp/pull/223#discussion_r3620114038. When this lands and https://github.com/webmachinelearning/webmcp/pull/223 is rebased on top of it, it'll be easier to make `RegisteredTool` expose an filled-out annotations object only when it was provided at registration time, and `undefined` otherwise.

CC @beaufortfrancois.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/225.html" title="Last updated on Jul 21, 2026, 3:26 PM UTC (16e587a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/225/1aece7c...16e587a.html" title="Last updated on Jul 21, 2026, 3:26 PM UTC (16e587a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -445,7 +445,8 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
    :: steps that invoke |tool|'s {{ModelContextTool/execute}}
 
    : [=tool definition/annotations=]
-   :: null if |tool|'s {{ModelContextTool/annotations}} does not [=map/exist=]. Otherwise, an [=annotations=] with the following [=struct/items=]:
+   :: null if |tool|'s {{ModelContextTool/annotations}} does not [=map/exist=]. Otherwise, an
+      [=annotations=] with the following [=struct/items=]:
 
       : [=annotations/read-only hint=]
       :: |tool|'s {{ModelContextTool/annotations}}'s {{ToolAnnotations/readOnlyHint}}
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -172,16 +172,24 @@ A <dfn>tool definition</dfn> is a [=struct=] with the following [=struct/items=]
      "internal" steps that have not been defined yet, that describe how to fill out a <{form}> and
      its [=form-associated elements=].
 
+  : <dfn>annotations</dfn>
+  :: an [=annotations=]-or-null.
```
beaufortfrancois, line 176: Nit: Shall we use `an optional [=annotations=]`?

beaufortfrancois (approved):
LGTM

index.bs
domfarolino: I think optional can be good for algorithm parameters, i.e., https://infra.spec.whatwg.org/#algorithm-params. But for members I think it's more common to see the type contain everything you need to know, including nullability. Structs like https://html.spec.whatwg.org/#specifier-resolution-record come to mind, whereas I don't think it's common (at least I can't point to any) to have structs with optional members. I think this is because fundamentally the member itself is not "optional"—it always exists, it's just whether its value contains some kind of sentinel null-like thing.

Issue #48, opened 2025-11-03
Add a callback to inform the site when tool invocation should be aborted


khushalsagar:
Since tool invocation is async, there will be cases when the user cancels the request which initiated the execution. We should likely have a callback which informs the site about this.

I don't think the other way around is necessary. If the site decides to abort, it can indicate so in the output of the execution. On that note, if we decide to go with https://github.com/webmachinelearning/webmcp/issues/9 then maybe we need defined error values. So the author can either return a value based on the output schema (success) or one of the error codes (failure/abort).

bwalderman:
Possibly a job for [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#implementing_an_abortable_api) which is already used for cancelling fetch requests.

Tool execute functions could take an optional `signal` parameter and listen for the `'abort'` event which they could use to roll back the operation and reset any page UI if needed.

The browser would need to, in effect, create an [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) on each tool call request and pass that AbortController's `.signal` object to the tool execute function. If the agent/user decides to cancel the request, then `AbortController.abort()` would be called.

khushalsagar:
AbortSignal does seem like the right primitive. There's an existing pattern on the `navigate` event where the browser passes a [signal](https://html.spec.whatwg.org/multipage/nav-history-apis.html#dom-navigateevent-signal) so script can listen to cancellations.

> Tool execute functions could take an optional signal parameter and listen for the 'abort' event which they could use to roll back the operation and reset any page UI if needed.

I'm not sure where the `signal` parameter should go. Why not have it on the `agent` interface for example? So we have a mental model of where other dependencies the site needs during tool execution should live.

anssiko:
>[RESOLUTION](https://www.w3.org/2026/01/22-webmachinelearning-minutes.html#65b6): Use AbortSignal to signal cancellation

beaufortfrancois:
Do I understand correctly that we're expecting the spec to be updated with something like `callback ToolExecuteCallback = Promise<any> (object input, ToolExecuteOptions options)` where `options` would be a dictionnary that contains an `AbortSignal signal` parameter to match the following web-platform test?

```js
    document.modelContext.registerTool({
      name: 'parent_tool',
      description: 'Parent tool',
      execute: async (input, opts) => {
        resolve_tool_started();

        // If an `AbortSignal` is not passed into the `execute()` callback, reject
[...]

Issue #218, opened 2026-07-01
Should unregistering a tool cancel an in-flight executions?


domfarolino:
How should tool unregistration interact with a tool execution that is already in-flight? This arises in two common patterns:

1. **Declarative Tools (React / SPA Form Detachment):**
   When a declarative form tool (`<form toolname="...">`) is submitted, the site calls [`e.respondWith(promise)`](https://github.com/webmachinelearning/webmcp/blob/main/declarative-api-explainer.md#events). After this method is called, apps built with frameworks like React might unmount the form element after submission, and render a success message instead. This removes the `<form>` from the DOM while `promise` is still pending. The plan is to make DOM removal trigger tool unregistration (what Chromium does now), which means we'd end up synchronously unregistering the tool before `promise` resolves, and before the invoker of the tool gets any response.
2. **Imperative Tools:**
   A tool caller invokes `modelContext.executeTool()` (soon to be spec'd) and gets a Promise representing the result. After tool call execution has started, but before the [`ToolExecuteCallback`](https://webmachinelearning.github.io/webmcp/#callbackdef-toolexecutecallback) Promise resolves, the site hosting the tool aborts the registration signal, thus unregistering the tool.

Right now in Chromium, unregistering a tool unconditionally aborts any in-flight executions. The result is a tricky footgun we're observing: **any declarative form using `e.respondWith()` will fail in React or SPAs that unmount forms immediately upon submission.**

### The question

Should we cancel pending executions for a tool when it is unregistered, or should we let them settle naturally? I personally think that tool unregistration should cancel pending invocations, to keep things simple, but I'm also not sure how big of a footgun this is for apps using frameworks.

We first observed this issue in sites using React. The pattern was something like:

```js
import { useState } from 'react';

export default function ToggleForm() {
  const [isSubmitted, setIsSubmitted] = useState(false);

  const handleSubmit = (e: React.SubmitEvent) => {
    e.preventDefault();
    if (e.nativeEvent.agentInvoked) {
      e.nativeEvent.respondWith('Your submission was successful');
    }
    setIsSubmitted(true);
  };

  // If submitted, show a success message instead of the form
  if (isSubmitted) {
    return <p>Your submission was successful and the form has been removed.</p>;
  }
```

Initially, our thought was just to gate the `setIsSubmitted(true);` behind a macrotask to let the Promise generated by `respondWith()` resolve first:

```diff
-    setIsSubmitted(true);
+     setTimeout(() => {
[...]

PR #226, opened 2026-07-21
Spec the `executeTool()` API

### LINKED ISSUE #51: Define the API for in-page Agents to use a site's declared tools

### LINKED ISSUE #57: Allow iframes to declare tools

### LINKED ISSUE #117: Allow iframes to deal with tools in their parent/top-level context

domfarolino:
This PR specifies the `executeTool()` API on `ModelContext`.

Closes https://github.com/webmachinelearning/webmcp/issues/51.
Closes https://github.com/webmachinelearning/webmcp/issues/57.
Closes https://github.com/webmachinelearning/webmcp/issues/117.
Closes https://github.com/webmachinelearning/webmcp/issues/159.
Closes https://github.com/webmachinelearning/webmcp/issues/160.

This PR also introduces the `AbortSignal` member to `ModelContextExecuteToolOptions`, but it is not wired up to actually abort the tool owner's callback yet. That will be handled in a follow-up PR towards fixing #48.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/226.html" title="Last updated on Aug 14, 2026, 3:06 PM UTC (ee98c1b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/226/5801678...ee98c1b.html" title="Last updated on Aug 14, 2026, 3:06 PM UTC (ee98c1b)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,7 +325,7 @@ By default, WebMCP is enabled in top-level `Window`s and its same-origin iframes
 
 Calls to `document.modelContext.registerTool()` will return a promise rejected with `NotAllowedError` DOMException when the permission is disabled, whether by the `allow` attribute or the `Permissions-Policy: tools=()` header. Handling of declarative tool registration errors, including when the permission is disabled is TBD; see [Issue #182](https://github.com/webmachinelearning/webmcp/issues/182).
 
-#### Cross-origin iframe exposure: `exposedTo`
+#### Cross-origin iframe exposure: `registerTool() and `exposedTo`
 
 By default, tools registered by a document are only exposed to itself, same-origin documents in the same tree, and built-in browser agents (see this <a href=#built-in-agent-default-exposure>discussion</a>). To support author-provided agents running in frames, developers can selectively share tools with secure origins of their choice, `exposedTo` option during registration:
 
@@ -342,9 +342,76 @@ Any document in the tree matching these origins (and allowed to use `tools` perm
 - Receive the `toolchange` event on its `document.modelContext` when the tool is registered or unregistered.
 - Be able to discover and run the tools
 
-#### Discovering and running tools
+#### Discovering and running tools: `getTools()` and `executeTool()`
 
-TODO: Spec and describe the `modelContext.getTools()` and `modelContext.executeTool()` APIs.
+Once tools are registered, in-page agents can discover and invoke them using `getTools()` and `executeTool()`.
+
+Calling `document.modelContext.getTools()` returns a promise that resolves with an array of `RegisteredTool` dictionary objects. Each object contains the tool's `name`, `description`, `inputSchema`, `origin`, and owner `window`:
+
+```js
+// Discover tools exposed by same-origin frames in the tree (default)
+const tools = await document.modelContext.getTools();
+
+for (const tool of tools) {
+  console.log(`Tool: ${tool.name} (from ${tool.origin})`);
+  console.log(`Description: ${tool.description}`);
+  console.log(`Parameters schema:`, tool.inputSchema);
+}
+
+// Or discover tools exposed by specific cross-origin partner frames:
+const crossOriginTools = await document.modelContext.getTools({
+  fromOrigins: ["https://trusted-partner.example"]
+});
+```
+
+An agent executes a discovered `RegisteredTool` by passing the tool dictionary and input arguments along to `document.modelContext.executeTool()`. The browser securely mediates the execution, ensuring the [`exposedTo`](https://webmachinelearning.github.io/webmcp/#dom-modelcontextregistertooloptions-exposedto) and [`fromOrigins`](https://webmachinelearning.github.io/webmcp/#dom-modelcontextgettooloptions-fromorigins) agree, and the tool runs in the tool owner's execution context:
+
+```js
+const tools = await document.modelContext.getTools();
+const addTodoTool = tools.find(t => t.name === "add-todo");
+
+if (addTodoTool) {
+  try {
+    const result = await document.modelContext.executeTool(
+      addTodoTool,
+      { text: "Buy groceries" }
+    );
+    console.log("Tool result:", result);
+  } catch (error) {
+    console.error("Tool execution failed:", error);
+  }
+}
+```
+
+##### Cancelling execution with `AbortSignal`
+
+Tool invocations can be cancelled mid-execution (e.g., if the user aborts an ongoing request, as they might with the "stop button" that's present in most agent UIs) by passing an `AbortSignal`:
+
+```js
+const controller = new AbortController();
+
+const executionPromise = document.modelContext.executeTool(
+  addTodoTool,
+  { text: "Buy groceries" },
+  { signal: controller.signal }
+);
+
+// If the user cancels the interaction:
+stopButton.addEventListener('click', e => controller.abort());
+```
+
+The tool's [execution callback](https://webmachinelearning.github.io/webmcp/#callbackdef-toolexecutecallback) receives this signal via its [`options.signal`](https://webmachinelearning.github.io/webmcp/#dom-modelcontextexecutetooloptions-signal) parameter, allowing it to abort underlying network requests or asynchronous tasks cleanly.
+
+##### Responding to dynamic tool updates: the `toolchange` event
+
+When tools are added, removed, or updated dynamically (such as when user interactions result in new tools being registered), `document.modelContext` fires a `toolchange` event:
+
+```js
+document.modelContext.addEventListener("toolchange", async () => {
+  const currentTools = await document.modelContext.getTools();
+  updateAgentToolRegistry(currentTools);
+});
+```
 
 
 ## Alternatives Considered
--- a/index.bs
+++ b/index.bs
@@ -193,6 +193,76 @@ An <dfn>annotations</dfn> is a [=struct=] with the following [=struct/items=]:
   :: a [=boolean=], initially false.
 </dl>
 
+<h3 id="pending-tool-executions">Pending tool executions</h3>
+
+A <dfn>pending tool execution</dfn> is a [=struct=] with the following [=struct/items=]:
+
+<dl dfn-for="pending tool execution">
+  : <dfn>caller document</dfn>
+  :: a {{Document}}.
+
+  : <dfn>target document</dfn>
+  :: a {{Document}}.
+
+  : <dfn>tool name</dfn>
+  :: a [=string=].
+
+  : <dfn>completion steps</dfn>
+  :: an algorithm that takes a [=string=]-or-null and a [=boolean=].
+</dl>
+
+A [=traversable navigable=] has a <dfn for="traversable navigable">pending tool executions
+map</dfn>, which is a [=map=] whose keys are [=unique internal values=] and whose values are
+[=pending tool execution=] [=structs=]. It is initially empty.
+
+Note: This map is only ever mutated from steps [=in parallel=]. This simulates the single,
+authoritative "browser process" that most modern browsers implement, where execution tracking sits
+outside any individual Document process's event loop, and is accessed asynchronously via some
+inter-process communication mechanism.
+
+<hr>
+
+<div algorithm=unloading-document-cleanup-steps>
+This specification's [=unloading document cleanup steps=], given a {{Document}} |document|, are as
+follows:
+
+1. Let |traversable| be |document|'s [=node navigable=]'s [=navigable/traversable navigable=].
+
+1. Run the following steps [=in parallel=]:
+
+   1. Let |executionsToRemove| be an empty [=list=].
+
+   1. [=map/For each=] |uuid| &rarr; |execution| of |traversable|'s [=traversable navigable/pending
+      tool executions map=]:
+
+      1. If |document| is |execution|'s [=pending tool execution/target document=] or |document| is
+         |execution|'s [=pending tool execution/caller document=], then [=list/append=] |uuid| to
+         |executionsToRemove|.
+
+   1. [=list/For each=] |uuid| of |executionsToRemove|:
+
+      1. Let |execution| be |traversable|'s [=traversable navigable/pending tool executions
+         map=][|uuid|].
+
+      1. <p id=caller-destroyed-cleanup>If |document| is |execution|'s [=pending tool
+         execution/target document=], then run |execution|'s [=pending tool execution/completion
+         steps=] given null and false.</p>
+
+         Note: This removes |execution| from the [=traversable navigable/pending tool executions
+         map=].
+
+      1. If |document| is |execution|'s [=pending tool execution/caller document=], then:
+
+         1. [=map/Remove=] |traversable|'s [=traversable navigable/pending tool executions
+            map=][|uuid|].
+
+         1. Issue(#48): Abort the `AbortSignal` in |execution|'s [=pending tool execution/target
+            document=], so the tool can abort early now that the caller is dead.
+
+      1. [=Assert=]: |traversable|'s [=traversable navigable/pending tool executions map=][|uuid|]
+         does not [=map/exist=].
+
+</div>
 
 <hr>
 
@@ -288,7 +358,7 @@ a [=string=] |inputArguments|, and an algorithm |completionSteps|, are as follow
     when <a href=https://github.com/webmachinelearning/webmcp/issues/92>issue #92</a> is
     resolved.</p>
 
-    <div class=example id=unregistration-race>
+    <div class=example id=unregistration-execution-race>
      <xmp class=language-js>
      // -- Tool owner document. --
      const oldInputSchema = {...};
@@ -371,10 +441,33 @@ To <dfn for="model context">unregister a tool</dfn> given a {{ModelContext}} |mo
 
 1. [=map/Remove=] |tool map|[|tool name|].
 
+1. Let |targetDocument| be |modelContext|'s [=relevant global object=]'s [=associated
+   Document|associated <code>Document</code>=].
+
+1. Let |traversable| be |targetDocument|'s [=node navigable=]'s [=navigable/traversable navigable=].
+
 1. Run the following steps [=in parallel=]:
 
-   1. [=Notify documents of a tool change=] given |modelContext|'s [=relevant global object=]'s
-      [=associated Document|associated <code>Document</code>=] and |exposed origins|.
+   1. Let |canceledExecutions| be an empty [=list=].
+
+   1. [=map/For each=] |uuid| &rarr; |execution| of |traversable|'s [=traversable navigable/pending
+      tool executions map=]:
+
+      1. If |execution|'s [=pending tool execution/target document=] is |targetDocument| and
+         |execution|'s [=pending tool execution/tool name=] is |tool name|, then [=list/append=]
+         |uuid| to |canceledExecutions|.
+
+   1. [=list/For each=] |uuid| of |canceledExecutions|:
+
+      1. Let |execution| be |traversable|'s [=traversable navigable/pending tool executions
+         map=][|uuid|].
+
+      1. Run |execution|'s [=pending tool execution/completion steps=] given null and false.
+
+         Note: This removes |execution| from the [=traversable navigable/pending tool executions
+         map=].
+
+   1. [=Notify documents of a tool change=] given |targetDocument| and |exposed origins|.
 
 </div>
 
@@ -789,13 +882,66 @@ The <dfn method for=ModelContext>executeTool(<var>tool</var>, <var>inputArgument
 
       Issue: Support more granular errors than "{{UnknownError}}", based on each failure case.
 
+   1. Let |uuid| be a new [=unique internal value=].
+
    1. Let |completionSteps| be an algorithm that takes a [=string=]-or-null |result| and a [=boolean=] |success|, and runs the following steps:
 
       1. [=Assert=]: these steps are running [=in parallel=].
 
+      1. If |targetDocument|'s [=node navigable=]'s [=navigable/traversable navigable=]'s [=traversable navigable/pending tool executions map=][|uuid|] does not [=map/exist=], then return.
+
+         <div class=note>
+          <p>It is possible that a pending execution identified by |uuid| no longer exists. This can
+          happen due to a race between (a) tool cancellation when the caller document <a
+          href=#caller-destroyed-cleanup>gets destroyed</a>; and (b) tool promise resolution. Both
+          of these race to invoke |completionSteps|, and the first invocation will remove the
+          pending execution by its key |uuid|, this check protects subsequent racing invocations.</p>
+
+          <p>This can also happen due to a limitation with tool unregistration. Consider this
+          example:</p>
+
+          <div class=example id=unregistration-race>
+           <xmp class=language-js>
+            const ac = new AbortController();
+            document.modelContext.registerTool({
+              name: "tool"
+              description: "a racy tool",
+              execute: async () => {
+                return new Promise(resolve => {
+                  resolve("tool result");
+                  ac.abort();
+                });
+              }
+            }, {signal: ac.signal});
+           </xmp>
+
+           <p>In this case, tool unregistration happens after a tool Promise is resolved, but before
+           its fulfillment handler runs. Since both run |completionSteps|, this check protects the
+           second invocation from trying to remove the same pending execution twice.</p>
+          </div>
+         </div>
+
+      1. [=map/Remove=] |targetDocument|'s [=node navigable=]'s [=navigable/traversable navigable=]'s [=traversable navigable/pending tool executions map=][|uuid|].
+
       1. If |success| is true, then [=queue a global task=] on the [=webmcp task source=] given |invokingDocument|'s [=relevant global object=] to [=resolve=] |promise| with |result|.
 
       1. Otherwise, [=queue a global task=] on the [=webmcp task source=] given |invokingDocument|'s [=relevant global object=] to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}.
+
+   1. Let |execution| be a new [=pending tool execution=], with the following [=struct/items=]:
+
+      : [=pending tool execution/caller document=]
+      :: |invokingDocument|
+
+      : [=pending tool execution/target document=]
+      :: |targetDocument|
+
+      : [=pending tool execution/tool name=]
+      :: |toolName|
+
+      : [=pending tool execution/completion steps=]
+      :: |completionSteps|
+
+   1. Set |targetDocument|'s [=node navigable=]'s [=navigable/traversable navigable=]'s [=traversable navigable/pending tool executions map=][|uuid|] to |execution|.
 
    1. [=Queue a global task=] on the [=webmcp task source=] given |targetWindow| to run the [=tool execute steps=] given |toolName|, |targetDocument|, |inputArguments|, and |completionSteps|.
 
@@ -1028,9 +1174,10 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
 </pre>
 
 <div algorithm>
-The <dfn>declarative execute steps</dfn> are as follows.
+The <dfn>declarative execute steps</dfn> are as follows:
 
-domfarolino
+Issue: Spec the declarative execution steps, and their integration with form elements.
+
 </div>
 
 <h3 id="events">Events</h3>
--- a/index.bs
+++ b/index.bs
@@ -417,7 +417,7 @@ The {{ModelContext}} interface provides methods for web applications to register
 interface ModelContext : EventTarget {
   Promise<undefined> registerTool(ModelContextTool tool, optional ModelContextRegisterToolOptions options = {});
   Promise<sequence<RegisteredTool>> getTools(optional ModelContextGetToolOptions options = {});
-  Promise<DOMString?> executeTool(RegisteredTool tool, DOMString inputArguments, optional ExecuteToolOptions options = {});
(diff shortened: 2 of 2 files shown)


index.bs
```diff
@@ -323,6 +354,7 @@ The {{ModelContext}} interface provides methods for web applications to register
 interface ModelContext : EventTarget {
   Promise<undefined> registerTool(ModelContextTool tool, optional ModelContextRegisterToolOptions options = {});
   Promise<sequence<RegisteredTool>> getTools(optional ModelContextGetToolOptions options = {});
+  Promise<DOMString?> executeTool(RegisteredTool tool, DOMString inputArguments, optional ExecuteToolOptions options = {});
```
beaufortfrancois: For consistency with the options naming convention, should we also prepend `ModelContext` to this dictionary name (e.g., `ModelContextExecuteToolOptions`)?
```suggestion
  Promise<DOMString?> executeTool(RegisteredTool tool, DOMString inputArguments, optional ModelContextExecuteToolOptions options = {});
```

I now wonder if we should have used `ModelContextRegisteredTool` as well ;)

index.bs
```diff
@@ -590,6 +629,107 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
...
+The <dfn method for=ModelContext>executeTool(<var>tool</var>, <var>inputArguments</var>,
+<var>options</var>)</dfn> method steps are:
+
+1. Let |invokingDocument| be [=this=]'s [=relevant global object=]'s [=associated
+   Document|associated <code>Document</code>=].
+
+1. If |invokingDocument| is not [=Document/fully active=], then return [=a promise rejected with=]
+   an "{{InvalidStateError}}" {{DOMException}}.
+
+1. If this's [=surrounding agent=]'s [=agent cluster=]'s [=is origin-keyed=] is false and this's
```
```suggestion
1. If [=this=]'s [=surrounding agent=]'s [=agent cluster=]'s [=is origin-keyed=] is false and [=this=]'s
```

This suggestion applies as well to getTools and registerTool.

index.bs
```diff
@@ -590,6 +629,107 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
...
+   "<code>file</code>", then return [=a promise rejected with=] a "{{SecurityError}}"
+   {{DOMException}}.
+
+1. If |invokingDocument| is not [=allowed to use=] the "{{tools}}" feature, then return [=a promise
+   rejected with=] a "{{NotAllowedError}}" {{DOMException}}.
+
+1. Let |expectedTargetOriginURL| be the result of [=URL parser|parsing=] |tool|'s
+   {{RegisteredTool/origin}}.
+
+1. If |expectedTargetOriginURL| is failure, then return [=a promise rejected with=] a
```
beaufortfrancois: How about using `NotSupportedError` instead and return a rejected promise in both cases?
- expectedTargetOriginURL is a failure
- expectedTargetOrigin is opaque

index.bs
```diff
@@ -590,6 +629,107 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
...
+
+1. Let |expectedTargetOrigin| be |expectedTargetOriginURL|'s [=url/origin=].
+
+1. [=Assert=]: |expectedTargetOrigin| is not an [=opaque origin=].
+
+1. Let |promise| be [=a new promise=] created in [=this=]'s [=relevant realm=].
+
+1. Let |signal| be |options|'s {{ExecuteToolOptions/signal}} if it [=map/exists=]; otherwise null.
+
+1. If |signal| is not null, then:
```
```suggestion
1. Let |signal| be |options|'s {{ExecuteToolOptions/signal}}.

1. If |signal| [=map/exists=], then:

File: index.bs [hunk 5]
```diff
@@ -590,6 +629,107 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
...
+
+1. Let |targetDocument| be |targetWindow|'s [=associated Document|associated
+   <code>Document</code>=].
+
+1. Run the following steps [=in parallel=]:
+
+   1. If |targetDocument|'s [=node navigable=]'s [=navigable/traversable navigable=] is not
+      |invokingDocument|'s [=node navigable=]'s [=navigable/traversable navigable=], then [=queue a
+      global task=] on the [=webmcp task source=] given |invokingDocument|'s [=relevant global
+      object=] to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these
```
beaufortfrancois, line 859: UnknownError feels weird to me in these cases. Maybe InvalidStateError?

index.bs
```diff
@@ -257,6 +256,38 @@ a [=list=] of [=origins=] |exposed origins|, and an [=origin=] |accessing origin
...
+
+1. [=promise/React=] to |toolPromise|:
+
+   - If |toolPromise| was fulfilled with value |v|:
+
+      1. Let |serializedResult| be the result of [=serializing a JavaScript value to a JSON string=] given |v|. If this throws an exception, run |completionSteps| given null and false, and abort these steps.
+
+      1. Run |completionSteps| given |serializedResult| and true.
+
+   - If |toolPromise| was rejected with reason <var ignore>r</var>, then run |completionSteps| given null and false.
```
beaufortfrancois: We may want to add that implementations may show a devtools warning that displays `r`, otherwise it's hard to know why `toolPromise` was rejected. This is what Chromium does at https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/script_tools/model_context.cc;l=293;drc=43b33e1c1959dd05d50dd707209464fda44222ed

index.bs
```diff
@@ -590,6 +629,107 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
...
+1. Let |signal| be |options|'s {{ExecuteToolOptions/signal}} if it [=map/exists=]; otherwise null.
+
+1. If |signal| is not null, then:
+
+   1. If |signal| is [=AbortSignal/aborted=], then return [=a promise rejected with=] |signal|'s
+      [=AbortSignal/abort reason=].
+
+   1. [=AbortSignal/add|Add the following abort steps=] to |signal|:
+
+      1. [=Reject=] |promise| with |signal|'s [=AbortSignal/abort reason=].
```
beaufortfrancois, line 842: We should also have an issue to spec and fire `toolcancel` event there. See https://github.com/webmachinelearning/webmcp/pull/146

index.bs
```diff
@@ -590,6 +629,107 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
...
+   1. If |targetToolMap|[|toolName|] does not [=map/exist=], then [=queue a global task=] on the
+      [=webmcp task source=] given |invokingDocument|'s [=relevant global object=] to [=reject=]
+      |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
+
+      Issue: Support more granular errors than "{{UnknownError}}", based on each failure case.
+
+   1. Let |tool definition| be |targetToolMap|[|toolName|].
+
+   1. If [=tool is exposed to an origin=] given |targetOrigin|, |tool definition|'s [=tool
+      definition/exposed origins=], and |callerOrigin| is false, then [=queue a global task=] on the
```
```suggestion
      definition/exposed origins=], and |callerOrigin| returns false, then [=queue a global task=] on the

File: index.bs [hunk 9]
```diff
@@ -590,6 +629,107 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
...
+      [=webmcp task source=] given |invokingDocument|'s [=relevant global object=] to [=reject=]
+      |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
+
+      Issue: Support more granular errors than "{{UnknownError}}", based on each failure case.
+
+   1. Let |completionSteps| be an algorithm that takes a [=string=]-or-null |result| and a [=boolean=] |success|, and runs the following steps:
```
beaufortfrancois: Nit: How about we switch params order?
```suggestion
   1. Let |completionSteps| be an algorithm that takes a [=boolean=] |success| and a [=string=]-or-null |result|, and runs the following steps:

File: index.bs (hunk 1 above)
Comment (beaufortfrancois) [inline index.bs]: Why is DOMString optional in this spec PR?

File: index.bs [hunk 10]
```diff
@@ -348,6 +380,12 @@ is a [=model context=] [=struct=] created alongside the {{ModelContext}}.
     agents written in JavaScript, and possibly living in <{iframe}>s. The [=user agent=]'s [=browser
     agent=] uses a different internal mechanism to retrieve the tools exposed to it.</p>
   </dd>
+
+  <dt><code><var ignore>document</var>.{{Document/modelContext}}.{{ModelContext/executeTool(tool, inputArguments, options)}}</code></dt>
+  <dd>
+    <p>Executes a tool on the document it was registered on. Returns a promise that resolves to the
+    result of the tool's execution.</p>
```
beaufortfrancois: Maybe extends that it's a stringified result only for now.

index.bs
```diff
@@ -590,6 +629,107 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
 
 </div>
 
+<div algorithm>
+The <dfn method for=ModelContext>executeTool(<var>tool</var>, <var>inputArguments</var>,
+<var>options</var>)</dfn> method steps are:
+
+1. Let |invokingDocument| be [=this=]'s [=relevant global object=]'s [=associated
```
beaufortfrancois: Nit: Shall we use `executingDocument`?

index.bs
```diff
@@ -590,6 +629,107 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
 
 </div>
 
+<div algorithm>
+The <dfn method for=ModelContext>executeTool(<var>tool</var>, <var>inputArguments</var>,
```
beaufortfrancois, line 803: A JS example of how to execute a tool from a cross-origin iframe may come handy in this spec. See previous comment at https://github.com/webmachinelearning/webmcp/pull/223#discussion_r3620127533

index.bs
```diff
@@ -257,6 +256,38 @@ a [=list=] of [=origins=] |exposed origins|, and an [=origin=] |accessing origin
...
 
+<div algorithm>
+The <dfn>imperative execute steps</dfn>, given a {{ModelContextTool}} |tool|, a {{Window}}
+|targetWindow|, a [=string=] |inputArguments|, and an algorithm |completionSteps|, are as follows:
+
+1. [=Assert=]: these steps are running on |targetWindow|'s [=relevant agent=]'s [=agent/event
+   loop=].
+
+1. Let |inputObject| be the result of [=parse a JSON string to a JavaScript value=] given
+   |inputArguments| and |targetWindow|'s [=relevant realm=]. If <a spec=webidl lt="an exception was thrown">exception was thrown</a>, then
```
beaufortfrancois: Why not add this webidl bit to `<pre class="link-defaults">` at the top of the file?

index.bs
beaufortfrancois: Out of scope for this PR.
 
Have we consider moving executeTool from document.modelContext to RegisteredTool interface hinted in https://issues.chromium.org/536063275? FWIW This would match what Puppeteer does: https://pptr.dev/guides/webmcp#executing-tools

beaufortfrancois (commented):
We should update README as well: https://github.com/webmachinelearning/webmcp#discovering-and-running-tools

index.bs
```diff
@@ -590,6 +629,107 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
...
+      Issue(#48): Wire up |signal| to the tool execution callback in the tool host's document, so that tool abort is communicated all the way through.
+
+1. Let |targetWindow| be |tool|'s {{RegisteredTool/window}}.
+
+1. Let |targetDocument| be |targetWindow|'s [=associated Document|associated
+   <code>Document</code>=].
+
+1. Run the following steps [=in parallel=]:
+
+   1. If |targetDocument|'s [=node navigable=]'s [=navigable/traversable navigable=] is not
```
bvandersloot-mozilla, line 856: Should we check that this document is fully active?

index.bs
```diff
@@ -590,6 +629,107 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
...
+      [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
+
+      Issue: Support more granular errors than "{{UnknownError}}", based on each failure case.
+
+   1. Let |targetToolMap| be |targetDocument|'s [=Document/associated ModelContext|associated
+      <code>ModelContext</code>=]'s [=ModelContext/internal context=]'s [=model context/tool map=].
+
+   1. Let |toolName| be |tool|'s {{RegisteredTool/name}}.
+
+   1. If |targetToolMap|[|toolName|] does not [=map/exist=], then [=queue a global task=] on the
```
bvandersloot-mozilla, line 882: Weird case we could get into here: `|targetToolMap|[|toolName|]` isn't the same tool as returned from `getTools`. It may have been modified in between.

domfarolino: Yep, some modifications are OK, others are not. I'll be carrying over some from our implementation before taking this PR out of draft mode.

domfarolino: OK this part should be covered about as deterministically as it can be. We not only check tool existence in the trusted "browser process" bits corresponding to the spec's parallel steps, but we also re-check existence them when we're back on the target document's event loop. If the tool doesn't exist by the time we make it there, then we error tool execution. This will abort the caller's promise.

https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/226.html#unregistration-race outlines a race that's still possible, not with tool existence, but tool "version", that I plan to consider addressing later with a possible new `Tool` interface. But we can save that discussion for then!

index.bs
domfarolino: I don't think so? Excluding existing-but-not-fully-active documents would only exclude documents in the bf-cache. But those event loops are already paused, [since a document doesn't process its tasks when it's non-fully-active](https://html.spec.whatwg.org/#event-loop-processing-model:fully-active).

So the intent here is to queue execution on the event loop of a document that is not fully active, for when it later becomes active, and out of the bf-cache. For script running *inside* the document, we definitely take a harsher approach (i.e., `executeTool()` [called _from_ a non-fully-active document](https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/226.html#ref-for-fully-active%E2%91%A1) definitely fails). But I think queueing for non-fully-active documents is OK.

In either case, I've added a note below the steps that queue the task on the target document, indicating that the document might not be fully active, and that the execution is queued for when it becomes fully active again, since I think we generally under-document this kind of scenario.

index.bs
domfarolino: Yeah I was torn between `Tool.execute(inputArgs)` vs `document.modelContext.executeTool(toolInterfaceObject)`. Was going to raise it to the working group after the summer "break". Will resolve this for now, but let's keep it in mind when we address that particular issue, because I think it's an important one.

domfarolino: Please comment something just like this on the crbug, and I'll make sure it gets surfaced in the working group discussion.

index.bs
domfarolino: Good call, done.

index.bs
domfarolino: Ah, I actually think `invokingDocument` is more clear here. `executingDocument` could mean the one that is actually executing the real tool (i.e., the tool host) OR the one that just triggered execution remotely, whereas I think `invokingDocument` clearly captures the former.

index.bs
domfarolino: I'm personally always a little bit torn on how to handle this, but I think I've settled on this heuristic which matches most spec code nicely:

 1. I almost _always_ see `[=map/exists=]` used with the array access operator on maps, or its equivalent IDL prose that we have here.
 2. If the result of direct array access yielded some kind of "entry" object or "iterator" whose "existence" was a member we could query (like `std::map::find` and `std::npos` in C++), then I'd feel comfortable grabbing it, and in a completely separate step, querying `[=map/exists=]`. Since it technically doesn't, I treat array access in specs how C++ treats array access (i.e., `operator[]` and `std::map::at()`), and treat existence checks and the handling of it in the null case, in the same line as a short-hand for what the impl actually does.

Given all of this, I've decided to land on basically what https://dom.spec.whatwg.org/#ref-for-map-exists does. We check existence, and if we're clear, unconditionally assign a variable in the nested steps. The double access is fine.

index.bs
domfarolino: Yeah that's reasonable. Let's do it for now, and unify everything else later to either keep or remove the prefix.

index.bs
domfarolino: Yeah I agree this is bad. I forgot to add my "Issue: Support more granular errors than "{{UnknownError}}", based on each failure case." disclaimer below this. I'm trying to just match the implementation for now, which doesn't support errors any more granular than this, so we can easily update the impl and spec at the same time when we change each error to be something more specific. I'll leave this as-is for now, until we start working on https://crbug.com/509555636. I added the disclaimer below this though.

index.bs
domfarolino: Hmm, from a pure spec perspective I don't feel strongly, but I'd like to keep this in sync with the implementation: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/content_extraction/script_tools.mojom;l=94;drc=305df0399efdb5dda74f474c92208f8a84f8678d for selfish reasons. If you want to submit a CL to re-order the impl, and a follow-up PR to bring the spec into alignment, I'd be supportive. But for now I will resolve.

index.bs
domfarolino: Nice, good catch. Done.

index.bs
domfarolino: Done. I realized the spec didn't cover the opaque origin case on its own, and would've failed its assert. Per https://crrev.com/c/8121241/comment/a263bec5_46b078a4/, I felt `"DataError"` was better, but I'm happy to use the more common `"NotSupportedError"` instead. I'll update the implementation.

domfarolino: https://crrev.com/c/8155899 for impl changes.

index.bs
domfarolino: Yeah, and we'll want to fire `toolcancel` in the target document, not the invoking document, so I add this to the note below. I think we can resolve this, and add full abort cancellation in a follow-up, corresponding to https://crrev.com/c/8025300.

index.bs
domfarolino: This was just to match the implementation. But from looking there, it seems we only take advantage of the nullness in the declarative-tool-navigates-cross-document case, and since declarative tools aren't spec'd yet, we can omit this for now and add it if we need to later, once we spec that. I'll make this non-null.

index.bs
beaufortfrancois: Nit: Let's not forget `this's` that also appears in getTools and registerTool algos

index.bs
beaufortfrancois: I was suggesting this change to be consistent with registerTool. We should update this as well to match this pattern (which does make sense by the way).

index.bs
beaufortfrancois: LGTM % `ontoolcanceled`. It should be `ontoolcancel` for now.

index.bs
beaufortfrancois: It was a nit so let's keep it like that for now.

index.bs
beaufortfrancois: Done in https://issues.chromium.org/issues/536063275#comment5

index.bs
```diff
@@ -950,6 +1006,12 @@ The <dfn>synthesize a declarative JSON Schema object algorithm</dfn>, given a <{
 }
 </pre>
 
+<div algorithm>
+The <dfn>declarative execute steps</dfn> are as follows.
+
+domfarolino
```
beaufortfrancois: Nit: Maybe replace with TODO for now

index.bs
```diff
@@ -590,6 +799,183 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
...
+      1. If |success| is true, then [=queue a global task=] on the [=webmcp task source=] given
+         |invokingDocument|'s [=relevant global object=] to [=resolve=] |promise| with |result|.
+
+      1. Otherwise, [=queue a global task=] on the [=webmcp task source=] given |invokingDocument|'s
+         [=relevant global object=] to [=reject=] |promise| with an "{{UnknownError}}"
+         {{DOMException}}.
+
+   1. Let |execution| be a new [=pending tool execution=], with the following [=struct/items=]:
+
+      : [=pending tool execution/caller document=]
```
beaufortfrancois, line 952: Instead of `caller document`, should we use `invoking document` to make it easier to follow along when reading different algorithms?

index.bs
```diff
@@ -590,6 +799,183 @@ The <dfn method for=ModelContext>getTools(<var>options</var>)</dfn> method steps
...
+              execute: async () => {
+                return new Promise(resolve => {
+                  resolve("tool result");
+                  ac.abort();
+                });
+              }
+            }, {signal: ac.signal});
+           </xmp>
+
+           <p>In this case, tool unregistration happens after a tool Promise is resolved, but before
```
beaufortfrancois, line 934: Do we have wpt for this case?

index.bs
```diff
@@ -273,10 +450,33 @@ To <dfn for="model context">unregister a tool</dfn> given a {{ModelContext}} |mo
...
+
+      1. Let |execution| be |traversable|'s [=traversable navigable/pending tool executions
+         map=][|uuid|].
+
+      1. Run |execution|'s [=pending tool execution/completion steps=] given null and false.
+
+         Note: This removes |execution| from the [=traversable navigable/pending tool executions
+         map=].
+
+   1. [=Notify documents of a tool change=] given |targetDocument| and |exposed origins|.
```
beaufortfrancois, line 479: Chromium implementation fires `toolchange` event before rejecting execution when aborted.
```js
const ac = new AbortController();
await document.modelContext.registerTool(
  { name: "tool", description: "A tool", execute: () => {} },
  { signal: ac.signal },
);

document.modelContext.ontoolchange = console.log("toolchange");

const [tool] = await document.modelContext.getTools();
document.modelContext
  .executeTool(tool, "{}")
  .catch(() => console.log("execution rejected"));
ac.abort();

// > toolchange
// > execution rejected
```

Both algorithms queue a global task on the webmcp task source. Are there things we can do implementation wise or in the spec to make sure ordering will stay the same across different implementations of WebMCP?

index.bs
domfarolino: Renamed the variable to `callerDocument` instead of the member to `invoking document`, to achieve the consistency you rightly wanted, but also chose *this* direction to align with the implementation, for selfish reasons :)

index.bs
domfarolino: I just added one in https://crrev.com/c/8025300. The expectations are pretty simple: basically it's just that unregistration wins (and the caller's Promise rejects), and the browser process doesn't "crash" because it tries to remove the same entry twice (i.e., the check is properly implemented).

index.bs
domfarolino: Hmm, your example is just invoking `console.log("toolchange")` synchronously which is why you're seeing this.

It should be:

```js
document.modelContext.ontoolchange = e => console.log("toolchange");
```

And then you'll observe what the spec and implementation both align on I think.

beaufortfrancois: My bad! You're right. Thanks!

index.bs
domfarolino: I added some examples to the explainer, since that's ultimately the best place for them, it's where most developers will go, and we've had a TODO explicitly there for getTools() and executeTool().

domfarolino:
FYI most of the reviews for this PR happened while in draft mode, graciously by @beaufortfrancois. So I think everything is all addressed and ready to merge here, including the documentation changes in the explainer.

(Note that related to this, Chromium recently [got support](https://crrev.com/c/8224887) for tools that keep executing after unregistration. The spec change for this will come after this PR, addressing https://github.com/webmachinelearning/webmcp/issues/218.)

README.md
```diff
@@ -342,9 +342,76 @@ Any document in the tree matching these origins (and allowed to use `tools` perm
...
+
+```js
+const tools = await document.modelContext.getTools();
+const addTodoTool = tools.find(t => t.name === "add-todo");
+
+if (addTodoTool) {
+  try {
+    const result = await document.modelContext.executeTool(
+      addTodoTool,
+      { text: "Buy groceries" }
```
beaufortfrancois, line 377: This line will be valid only when https://github.com/webmachinelearning/webmcp/issues/243 is addressed.

README.md
```diff
@@ -342,9 +342,76 @@ Any document in the tree matching these origins (and allowed to use `tools` perm
...
+##### Cancelling execution with `AbortSignal`
+
+Tool invocations can be cancelled mid-execution (e.g., if the user aborts an ongoing request, as they might with the "stop button" that's present in most agent UIs) by passing an `AbortSignal`:
+
+```js
+const controller = new AbortController();
+
+const executionPromise = document.modelContext.executeTool(
+  addTodoTool,
+  { text: "Buy groceries" },
```
beaufortfrancois, line 395: Same here

README.md
```diff
@@ -342,9 +342,76 @@ Any document in the tree matching these origins (and allowed to use `tools` perm
...
+// Discover tools exposed by same-origin frames in the tree (default)
+const tools = await document.modelContext.getTools();
+
+for (const tool of tools) {
+  console.log(`Tool: ${tool.name} (from ${tool.origin})`);
+  console.log(`Description: ${tool.description}`);
+  console.log(`Parameters schema:`, tool.inputSchema);
+}
+
+// Or discover tools exposed by specific cross-origin partner frames:
```
beaufortfrancois, line 361: Result actually includes same-origin frames in the tree + cross-origin tools. We should make this clear. See example at https://developer.chrome.com/docs/ai/webmcp/imperative-api#discover-tools

README.md
```diff
@@ -342,9 +342,76 @@ Any document in the tree matching these origins (and allowed to use `tools` perm
...
+  addTodoTool,
+  { text: "Buy groceries" },
+  { signal: controller.signal }
+);
+
+// If the user cancels the interaction:
+stopButton.addEventListener('click', e => controller.abort());
+```
+
+The tool's [execution callback](https://webmachinelearning.github.io/webmcp/#callbackdef-toolexecutecallback) receives this signal via its [`options.signal`](https://webmachinelearning.github.io/webmcp/#dom-modelcontextexecutetooloptions-signal) parameter, allowing it to abort underlying network requests or asynchronous tasks cleanly.
```
beaufortfrancois, line 403: https://webmachinelearning.github.io/webmcp/#issue-8efd9136 is not spec'ed yet right?

README.md
domfarolino: True, this example jumps ahead a bit. But I think that's OK if we fast-follow.

README.md
domfarolino: Correct, but I've almost got https://github.com/webmachinelearning/webmcp/compare/abortable-tools?expand=1 up and ready. So yeah these are a little de-sync'd but I think aspirational docs are OK hehe.

README.md
domfarolino: Agreed. I stole some of the wording from there and slotted it in a little more explicitly here, this commit: https://github.com/webmachinelearning/webmcp/commit/7abca24665c438a47e6770a3d1b470ca5e965764.

PR #240, opened 2026-08-11
Validate `exposedTo` before registering `AbortSignal` algorithm

beaufortfrancois:
Move the validation of `options.exposedTo` origins before attaching the unregister algorithm to `options.signal` in `registerTool()`.

Previously, the abort algorithm was attached to the signal before origin validation. If `exposedTo` contained an invalid or untrustworthy origin, the promise was rejected with a `SecurityError`, but the abort algorithm remained attached to the signal and could inadvertently unregister a subsequently registered tool of the same name upon aborting.

Also reorder `ModelContextRegisterToolOptions` dictionary members and domintro definitions to match the processing order of options.

Fixed in Chromium: https://chromium-review.googlesource.com/c/chromium/src/+/8236351

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/webmcp/pull/240.html" title="Last updated on Aug 14, 2026, 2:46 PM UTC (00a25c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/240/3678f64...beaufortfrancois:00a25c7.html" title="Last updated on Aug 14, 2026, 2:46 PM UTC (00a25c7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -407,6 +407,20 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
      </ol>
    </div>
 
+1. Let |exposed origins| be an empty [=list=] of [=origins=].
+
+1. If |options|'s {{ModelContextRegisterToolOptions/exposedTo}} [=map/exists=], then:
+
+     1. [=list/For each=] |origin| of |options|'s {{ModelContextRegisterToolOptions/exposedTo}}:
+
+          1. Let |parsedURL| be the result of running the [=URL parser=] on |origin|.
+
+          1. If |parsedURL| is failure or its [=url/origin=] is not [$is origin potentially
+             trustworthy?|potentially trustworthy$], then return [=a promise rejected with=] a
+             "{{SecurityError}}" {{DOMException}}.
+
+          1. [=list/Append=] |parsedURL|'s [=url/origin=] to |exposed origins|.
+
 1. Let |promise| be [=a new promise=] created in [=this=]'s [=relevant realm=].
 
 1. If |options|'s {{ModelContextRegisterToolOptions/signal}} [=map/exists=], then:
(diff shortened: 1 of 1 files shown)


domfarolino:
So the only observable with this change is the buggy ability to unregister a later-registered, valid tool by the abort algorithm hanging around, right?

(I think this will be fixed regardless, once we end up with a UUID-keyed tool registry and `Tool` interface, but this is probably fine in the meantime).

beaufortfrancois:
> So the only observable with this change is the buggy ability to unregister a later-registered, valid tool by the abort algorithm hanging around, right?

Yes.

index.bs
markafoltz, line 434: The Blink implementation checks this before checking `exposedTo` origins.

domfarolino: OK I think I fixed this in the latest commit.

domfarolino (approved):

index.bs
```diff
@@ -407,20 +407,9 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
...
-
-   1. If |signal| is [=AbortSignal/aborted=], then return [=a promise rejected with=]
-      |signal|'s [=AbortSignal/abort reason=].
-
-   1. [=AbortSignal/add|Add the following abort steps=] to |signal|:
-
-        1. [=model context/Unregister a tool=] given [=this=] and |tool name|.
-
-        1. [=Reject=] |promise| with |signal|'s [=AbortSignal/abort reason=].
+1. If |options|'s {{ModelContextRegisterToolOptions/signal}} [=map/exists=] and is
```
beaufortfrancois, line 410: I'd argue we could make it nicer by moving this below the `exposed origins` check so that we check first all things related to `exposed origins`, then and only then `signal` aborted + registering algo. Here, we're mixing both.
I'm happy to send a spec PR and Chromium CL if accepted.

domfarolino: I think while this is a little strange, we should avoid doing as much work as possible (aside from environment checks like at the very beginning of this method) if the developer passes in an aborted signal. IMO if the signal is already aborted, the Promise shouldn't abort due to any _other_ invalid input (that WebIDL doesn't already catch), since we shouldn't even process them. If you can think of a way to preserve that invariant while rearranging this, I'm all ears. Otherwise, I think the semantics here are at least worth preserving.

PR #244, opened 2026-08-16
Fix normative defects in tool registration, notification, and observation

### LINKED ISSUE #213: registerTool() stores the tool definition in the model context struct itself, not its tool map
In the registerTool() algorithm, the final insertion step writes the tool definition by indexing the internal context directly:

[index.bs:453](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L453)

Set this's internal context[|tool name|] to |tool definition|.
But a ModelContext's [internal context](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L321) is a model context struct, and a [model context](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L125) struct's only item is a tool map. Tools are meant to be keyed inside that tool map, not on the struct itself. Every other part of the spec reads/writes through the tool map:

The duplicate-name check earlier in the same algorithm: Let |tool map| be this's internal context's tool map … If |tool map|[|tool name|] exists ([index.bs:354](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L354), [:360](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L360))
[unregister a tool](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L258): operates on internal context's tool map
[perform an observation](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L680): reads internal context's tool map's values
Impact: A successfully registered tool is written to a location that unregister a tool and perform an observation never read, so per the literal spec text a registered tool would never be enumerated to agents and could never be unregistered. (Note the duplicate-name check at :360 reads the real tool map, so it also wouldn't observe tools inserted by this step — registering the same name twice would not be detected.)

Suggested fix: the |tool map| local is already in scope from :354; reuse it.

Set |tool map|[|tool name|] to |tool definition|.

Introduced in [#179](https://github.com/webmachinelearning/webmcp/pull/179).

### LINKED ISSUE #214: perform an observation asserts the active document is not fully active (inverted condition)
The [perform an observation](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L660) algorithm's second step asserts:

[index.bs:666](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L666)

Assert: |traversable|'s active document is not fully active.

This appears inverted. A [top-level traversable](https://html.spec.whatwg.org/#top-level-traversable)'s active document is always [fully active](https://html.spec.whatwg.org/#fully-active) per HTML, and observation is meant to run against live pages currently in view (the surrounding prose says implementations reserve it "for when the user is interacting with a browser agent while web content is in view"). As written, the assertion would fail for exactly the documents you intend to observe.

For contrast, registerTool() rejects when the document is not fully active ([index.bs:343](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L343)), which confirms the intended polarity for an active, observable document.

Suggested fix:

Assert: |traversable|'s active document is fully active.
Introduced in [#164](https://github.com/webmachinelearning/webmcp/pull/164).

### LINKED ISSUE #215: notify documents of a tool change skips the registering document (descendant navigables excludes it), contradicting the worked example
[notify documents of a tool change](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L186) computes the set of documents to notify as:

[index.bs:192](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L192)

Let |navigablesToNotify| be |tool owner|'s node navigable's traversable navigable's descendant navigables.
[descendant navigables](https://html.spec.whatwg.org/#descendant-navigables) does not include the document's own navigable, so the top-level / tool-owning document is never iterated and never receives toolchange.

This contradicts the spec's own example block immediately below the algorithm:

[index.bs:213–:232](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L213)

document.modelContext.ontoolchange = e => console.log('Parent toolchange');
iframe.contentDocument.modelContext.ontoolchange = e => console.log('Child toolchange');

const p = document.modelContext.registerTool({ /* ... */ });
// `Parent toolchange` will always log before `Child toolchange` ...
The example asserts the registering (parent) document logs Parent toolchange, but with descendant navigables the parent is excluded and would never fire. The PR that introduced this algorithm ([#179](https://github.com/webmachinelearning/webmcp/pull/179)) describes its intent as firing toolchange "in all documents in the tree that a tool changes in" — which includes the tool owner.

It's also inconsistent with [perform an observation](https://github.com/webmachinelearning/webmcp/blob/main/index.bs#L670), which uses [inclusive descendant navigables](https://html.spec.whatwg.org/#inclusive-descendant-navigables) (self-inclusive) for the equivalent traversal.

Suggested fix: use the self-inclusive variant.

Let |navigablesToNotify| be |tool owner|'s node navigable's traversable navigable's active document's inclusive descendant navigables.
Introduced in [#179](https://github.com/webmachinelearning/webmcp/pull/179)

emecii:
Three normative fixes, each a case where the algorithm text does not do what the surrounding spec says it does, plus one editorial nit in the same area.

### `registerTool()` writes to the struct instead of its tool map

Fixes #213

A [=model context=] struct has exactly one item, `tool map`, so indexing the struct itself by tool name is not well-defined. Every other algorithm reads through `internal context`'s `tool map` — see `unregister a tool`, `getTools()`, and `perform an observation`.

### `perform an observation` has an inverted assertion

Fixes #214

It asserts the traversable's active document is *not* fully active. A top-level traversable's active document is always fully active, and the algorithm then goes on to read that document's descendants, which only makes sense when it is. `registerTool()`, `getTools()`, and `executeTool()` all reject when a document is *not* fully active.

### `notify documents of a tool change` skips the registering document

Fixes #215

Using `descendant navigables` excludes the tool owner's own navigable, so the document that registered the tool never receives `toolchange`. This contradicts the worked example immediately below the algorithm, which asserts that `Parent toolchange` always logs before `Child toolchange` for a `registerTool()` call on the top-level document. The two comparable algorithms — in `getTools()` and `perform an observation` — both use `inclusive descendant navigables`.

### Editorial

Removes a stray possessive in `perform an observation` (`|descendant|'s active document's.`), matching the identical assignments in `notify documents of a tool change` and `getTools()`.

---

All three normative issues were filed by @Aravindargutus. No new link targets are introduced — every construct used already appears elsewhere in `index.bs`.

***
<a href="https://pr-preview.s3.amazonaws.com/emecii/webmcp/pull/244.html" title="Last updated on Aug 16, 2026, 4:54 AM UTC (f2ef15e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/244/384e110...emecii:f2ef15e.html" title="Last updated on Aug 16, 2026, 4:54 AM UTC (f2ef15e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -275,7 +275,7 @@ To <dfn>notify documents of a tool change</dfn> given a {{Document}} |tool owner
 1. [=Assert=]: these steps are running [=in parallel=].
 
 1. Let |navigablesToNotify| be |tool owner|'s [=node navigable=]'s [=navigable/traversable
-   navigable=]'s [=Document/descendant navigables=].
+   navigable=]'s [=Document/inclusive descendant navigables=].
 
 1. [=list/For each=] |navigable| of |navigablesToNotify|:
 
@@ -679,7 +679,8 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
    : [=tool definition/exposed origins=]
    :: |exposed origins|
 
-1. Set [=this=]'s [=ModelContext/internal context=][|tool name|] to |tool definition|.
+1. Set [=this=]'s [=ModelContext/internal context=]'s [=model context/tool map=][|tool name|] to
+   |tool definition|.
 
 1. Run the following steps [=in parallel=]:
 
@@ -1280,7 +1281,7 @@ steps:
 
 1. [=Assert=]: This algorithm is running in the [=browser agent=]'s [=AI agent queue=].
 
-1. [=Assert=]: |traversable|'s [=navigable/active document=] is not [=Document/fully active=].
+1. [=Assert=]: |traversable|'s [=navigable/active document=] is [=Document/fully active=].
 
 1. Let |observation| be a new [=observation=].
 
@@ -1289,7 +1290,7 @@ steps:
 
 1. [=list/For each=] [=navigable=] |descendant| of |flat descendants|:
 
-     1. Let |document| be |descendant|'s [=navigable/active document=]'s.
+     1. Let |document| be |descendant|'s [=navigable/active document=].
 
      1. Let |id| be |document|'s [=Document/unique ID=].
 


domfarolino (approved):
Thanks for catching these tricky issues, much appreciated!

PR #247, opened 2026-08-17
Spec `AbortSignal` integration with tool execution

### LINKED ISSUE #48: Add a callback to inform the site when tool invocation should be aborted

domfarolino:
This PR specifies abortable tools, via `AbortSignal` integration with `executeTool()`. Specifically, we introduce a new "local" pending execution map that is only ever mutated on the tool owner/target document's event loop, so it can track pending executions from its side, separately from the "traversable"-global map that's mutated in parallel and matches a canonical "browser process". We introduce the `ToolExecuteCallbackOptions` dictionary and its sole `signal` member, which is always passed into the [`ToolExecuteCallback`](https://webmachinelearning.github.io/webmcp/#callbackdef-toolexecutecallback) as its second argument.

This PR also handles various race conditions with tool execution, tool cancellation, and tool promise resolution/rejection, that can all happen across agent clusters. Each race condition is accounted for and documented, sometimes with examples.

Fixes #48.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/247.html" title="Last updated on Aug 18, 2026, 7:14 PM UTC (3bab858)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/247/9c7ce3e...3bab858.html" title="Last updated on Aug 18, 2026, 7:14 PM UTC (3bab858)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -963,7 +963,7 @@ The <dfn method for=ModelContext>executeTool(<var>tool</var>, <var>inputObject</
 
       1. Run the following steps [=in parallel=]:
 
-         1. [=Cancel a pending tool execution|cancel=] given |traversable| and |uuid|.
+         1. [=Cancel a pending tool execution=] given |traversable| and |uuid|.
 
 1. Run the following steps [=in parallel=]:
 
--- a/index.bs
+++ b/index.bs
@@ -136,8 +136,15 @@ A <dfn>model context</dfn> is a [=struct=] with the following [=struct/items=]:
      [=structs=].
 
   : <dfn>local pending tool executions map</dfn>
-  :: a [=map=] whose [=map/keys=] are [=unique internal values=] and whose [=map/values=] are [=local
-     pending tool execution=] [=structs=]. It is initially empty.
+  :: a [=map=] whose [=map/keys=] are [=unique internal values=] and whose [=map/values=] are
+     [=local pending tool execution=] [=structs=]. It is initially empty.
+
+     Note: this map is similar to a [=traversable navigable=]'s [=traversable navigable/pending tool
+     executions map=], but it only contains pending execution information for tools under a single
+     {{ModelContext}} object. It is used to store objects that can only be accessed from that
+     object's event loop, and because it is event-loop-local, it can get out of sync from the
+     [=traversable navigable=]'s more "global" map.
+
 </dl>
 
 A <dfn>tool definition</dfn> is a [=struct=] with the following [=struct/items=]:
--- a/index.bs
+++ b/index.bs
@@ -523,6 +523,8 @@ internal value=] |uuid|, are as follows:
 
    - If |toolPromise| was rejected with reason |r|, then:
 
+       1. Optionally [=report a warning to the console=] describing |r|.
+
        1. Let |localExecutions| be |targetDocument|'s [=Document/associated ModelContext|associated
           <code>ModelContext</code>=]'s [=ModelContext/internal context=]'s [=model context/local
           pending tool executions map=].
@@ -530,8 +532,6 @@ internal value=] |uuid|, are as follows:
        1. If |localExecutions|[|uuid|] does not [=map/exist=], then return.
 
        1. [=map/Remove=] |localExecutions|[|uuid|].
-
-       1. Optionally [=report a warning to the console=] describing |r|.
 
        1. Run |completionSteps| given null and false.
 
--- a/index.bs
+++ b/index.bs
@@ -238,8 +238,8 @@ outside any individual Document process's event loop, and is accessed asynchrono
 inter-process communication mechanism.
 
 <div algorithm>
-To <dfn for="pending tool execution">cancel a pending tool execution</dfn> given a [=traversable
-navigable=] |traversable| and a [=unique internal value=] |uuid|:
+To <dfn>cancel a pending tool execution</dfn> given a [=traversable navigable=] |traversable| and a
+[=unique internal value=] |uuid|:
 
 1. [=Assert=]: these steps are running [=in parallel=].
 
@@ -309,9 +309,8 @@ follows:
          Note: This removes |execution| from the [=traversable navigable/pending tool executions
          map=].
 
-      1. If |document| is |execution|'s [=pending tool execution/caller document=], then
-         [=pending tool execution/cancel a pending tool execution|cancel=] given |traversable| and
-         |uuid|.
+      1. If |document| is |execution|'s [=pending tool execution/caller document=], then [=cancel a
+         pending tool execution=] given |traversable| and |uuid|.
 
       1. [=Assert=]: |traversable|'s [=traversable navigable/pending tool executions map=][|uuid|]
          does not [=map/exist=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -134,6 +134,17 @@ A <dfn>model context</dfn> is a [=struct=] with the following [=struct/items=]:
   : <dfn>tool map</dfn>
   :: a [=map=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=tool definition=]
      [=structs=].
+
+  : <dfn>pending executions map</dfn>
+  :: a [=map=] whose [=map/keys=] are [=unique internal values=] and whose [=map/values=] are [=local
+     pending tool execution=] [=structs=]. It is initially empty.
+</dl>
+
+A <dfn>local pending tool execution</dfn> is a [=struct=] with the following [=struct/items=]:
```
beaufortfrancois: Nit: I would have moved "local pending tool execution" below "tool definition" struct to match "tool map" + "pending executions map" order

index.bs
```diff
@@ -134,6 +134,17 @@ A <dfn>model context</dfn> is a [=struct=] with the following [=struct/items=]:
   : <dfn>tool map</dfn>
   :: a [=map=] whose [=map/keys=] are [=strings=] and whose [=map/values=] are [=tool definition=]
      [=structs=].
+
+  : <dfn>pending executions map</dfn>
```
beaufortfrancois: `pending executions map` from ModelContext and `pending tool executions map` from traversable navigable look very similar at a first glance but serve different purposes. A small note explaining the difference may help future readers.

index.bs
```diff
@@ -1060,6 +1165,15 @@ The {{ToolAnnotations}} dictionary provides optional metadata about a tool:
...
 </dl>
 
+<h4 id="tool-execute-callback-options">ToolExecuteCallbackOptions Dictionary</h4>
+
+The {{ToolExecuteCallbackOptions}} dictionary carries options passed to a tool's
+{{ToolExecuteCallback}} when the tool is executed.
+
+<dl class="domintro">
+  : <code><var ignore>options</var>["{{ToolExecuteCallbackOptions/signal}}"]</code>
+  :: An {{AbortSignal}} that communicates when the execution of the tool has been aborted.
```
```suggestion
  :: An {{AbortSignal}} that communicates when the execution of the tool has been cancelled.
```

index.bs
```diff
@@ -169,8 +180,9 @@ A <dfn>tool definition</dfn> is a [=struct=] with the following [=struct/items=]
      [[!JSON-SCHEMA]]
 
   : <dfn>execute steps</dfn>
```
beaufortfrancois, line 181: https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/247.html#tool-definition-execute-steps also needs to be updated with uuid

index.bs
```diff
@@ -246,20 +305,20 @@ follows:
...
          Note: This removes |execution| from the [=traversable navigable/pending tool executions
          map=].
 
-      1. If |document| is |execution|'s [=pending tool execution/caller document=], then:
-
-         1. [=map/Remove=] |traversable|'s [=traversable navigable/pending tool executions
-            map=][|uuid|].
+      1. <p id=caller-destroyed-cleanup>Otherwise, if |document| is |execution|'s [=pending tool
+         execution/caller document=] and is not |execution|'s [=pending tool execution/target
+         document=], then [=cancel a pending tool execution=] given |traversable| and |uuid|.</p>
```
beaufortfrancois, line 324: Shall we do this in parallel?

index.bs
```diff
@@ -386,7 +445,8 @@ The <dfn>tool execute steps</dfn>, given a [=string=] |toolName|, a {{Document}}
 
 1. Let |tool| be |toolMap|[|toolName|].
 
-1. Run |tool|'s [=tool definition/execute steps=] given |inputArguments| and |completionSteps|.
+1. Run |tool|'s [=tool definition/execute steps=] given |inputArguments|, |completionSteps|,
+   |uuid|, and |targetDocument|.
```
beaufortfrancois: It's quite hard to follow how `|targetDocument|` is used. Would it help removing the declarative execute steps for now?

index.bs
```diff
@@ -428,6 +518,14 @@ The <dfn>imperative execute steps</dfn>, given a {{ModelContextTool}} |tool|, a
...
 
+       1. Let |localExecutions| be |targetDocument|'s [=Document/associated ModelContext|associated
+          <code>ModelContext</code>=]'s [=ModelContext/internal context=]'s [=model context/pending
+          executions map=].
+
+       1. If |localExecutions|[|uuid|] does not [=map/exist=], then return.
+
+       1. [=map/Remove=] |localExecutions|[|uuid|].
+
        1. Optionally [=report a warning to the console=] describing |r|.
```
beaufortfrancois: It seems like we're doing it earlier according to https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/script_tools/model_context.cc;l=313;drc=e5861528c1070eab8e706d7a7c1b42dac502f6a1

beaufortfrancois (changes requested):

index.bs
domfarolino: Sounds good.

index.bs
domfarolino: You mean the usage of it in registerTool()?

index.bs
domfarolino: Done, and I also converged the names a bit.

index.bs
domfarolino: I think you're right. Looks like the warning happens before removal, so I've moved this up as the first substep here.

index.bs
domfarolino: This whole algorithm is run in parallel at the very top, so I think we're good.

index.bs
domfarolino: I'm now realizing that the signatures and calls to a lot of these invoke algorithms don't match even on `main`. Let me fix this off of this branch and I'll make any more necessary adjustments here after.

index.bs
domfarolino: Can you clarify your concern? I think it's clear in https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/247.html#imperative-execute-steps how `targetDocument` is used (for map retrieval and the realm etc.), but I guess you're suggesting that it would be even clearer if we removed one layer of abstraction—the intermediate `[=tool definition/execute steps=]`—which just act as a pass-through to fork to either imperative/declarative? I guess it's true that fewer layers of abstraction make things more clear, but I'm not sure if this PR makes things worse, or if I'm misunderstanding your comment.

beaufortfrancois: https://github.com/webmachinelearning/webmcp/pull/250 actually resolved my issue. `|targetDocument|` was missing in some parts of the algorithm hence my confusion.

beaufortfrancois (approved):

PR #248, opened 2026-08-18
Preserve in-flight executions after unregistration

### LINKED ISSUE #218: Should unregistering a tool cancel an in-flight executions?

domfarolino:
This PR preserves in-flight tool executions after a tool has been unregistered, thus fixing #218.

Fixes: #218.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/pull/248.html" title="Last updated on Aug 19, 2026, 6:04 PM UTC (3304b21)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/248/3bab858...3304b21.html" title="Last updated on Aug 19, 2026, 6:04 PM UTC (3304b21)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -556,9 +556,8 @@ To <dfn for="model context">unregister a tool</dfn> given a {{ModelContext}} |mo
 1. Let |targetDocument| be |modelContext|'s [=relevant global object=]'s [=associated
    Document|associated <code>Document</code>=].
 
-1. Run the following steps [=in parallel=]:
-
-   1. [=Notify documents of a tool change=] given |targetDocument| and |exposed origins|.
+1. [=In parallel=], [=notify documents of a tool change=] given |targetDocument| and |exposed
+   origins|.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -249,7 +249,7 @@ To <dfn>cancel a pending tool execution</dfn> given a [=traversable navigable=] 
 1. If |traversable|'s [=traversable navigable/pending tool executions map=][|uuid|] does not
    [=map/exist=], then return.
 
-   Note: See <a href=#unregistration-race>this note</a> to learn how a tool's natural
+   Note: See <a href=#pending-execution-removal-race>this note</a> to learn how a tool's natural
    resolution/rejection can race with the caller's cancellation. This might result in the pending
    execution entry for |uuid| being removed before we get here. In that case, the
    {{ModelContext/executeTool()}} promise will still be rejected with the abort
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -556,29 +556,8 @@ To <dfn for="model context">unregister a tool</dfn> given a {{ModelContext}} |mo
...
-
-      1. Let |execution| be |traversable|'s [=traversable navigable/pending tool executions
-         map=][|uuid|].
-
-      1. Run |execution|'s [=pending tool execution/completion steps=] given null and false.
-
-         Note: This removes |execution| from the [=traversable navigable/pending tool executions
-         map=].
-
    1. [=Notify documents of a tool change=] given |targetDocument| and |exposed origins|.
```
```suggestion
1. [=In parallel=], [=Notify documents of a tool change=] given |targetDocument| and |exposed origins|.
```

beaufortfrancois (approved):
LGTM with nit

index.bs
domfarolino: Done (locally).

PR #251, opened 2026-08-20
Change `executeTool()` inputObject type from `object` to `any`

beaufortfrancois:
Following Blink IDL bindings folks feedback in https://chromium-review.googlesource.com/c/chromium/src/+/8250826, we're changing `executeTool()` inputObject type from `object` to `any` with special handling after https://github.com/webmachinelearning/webmcp/pull/246.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/webmcp/pull/251.html" title="Last updated on Sep 10, 2026, 4:25 PM UTC (b755b64)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/251/50c4b7f...beaufortfrancois:b755b64.html" title="Last updated on Sep 10, 2026, 4:25 PM UTC (b755b64)">Diff</a>


beaufortfrancois:
@domfarolino If that looks good to you, I'll merge https://chromium-review.googlesource.com/c/chromium/src/+/8250826

index.bs
```diff
@@ -604,7 +604,7 @@ The {{ModelContext}} interface provides methods for web applications to register
 interface ModelContext : EventTarget {
   Promise<undefined> registerTool(ModelContextTool tool, optional ModelContextRegisterToolOptions options = {});
   Promise<sequence<RegisteredTool>> getTools(optional ModelContextGetToolOptions options = {});
-  Promise<DOMString> executeTool(RegisteredTool tool, optional object inputObject = {}, optional ModelContextExecuteToolOptions options = {});
+  Promise<DOMString> executeTool(RegisteredTool tool, optional any inputObject, optional ModelContextExecuteToolOptions options = {});
```
domfarolino, line 611: Per the CL discussion, I'm wondering if this needs to be `optional` at all? I think `any` can accept `undefined` on its own, which might simplify things here?

index.bs
```diff
@@ -919,8 +919,16 @@ The <dfn method for=ModelContext>executeTool(<var>tool</var>, <var>inputObject</
 
 1. [=Assert=]: |expectedTargetOrigin| is not an [=opaque origin=].
 
-1. Let |inputArguments| be the result of [=serializing a JavaScript value to a JSON string=] given
-   |inputObject|. If this threw an exception, then return [=a promise rejected with=] that exception.
+1. Let |inputArguments| be "<code>{}</code>".
+
+1. If |inputObject| is not {{undefined}}, then:
```
domfarolino: Do we need this? The fallback case always gives us undefined (perhaps regardless of which way the above comment goes), which makes me think we can just check if it is an object type directly, and TC39 will handle it correctly?

index.bs
```diff
@@ -919,8 +919,16 @@ The <dfn method for=ModelContext>executeTool(<var>tool</var>, <var>inputObject</
...
+
+   1. If |inputObject| [=Object type|is not an Object=], then return [=a promise rejected with=] a
```
domfarolino: I think you want the Object check that step 1 in https://webidl.spec.whatwg.org/#js-object has, as opposed to this current one which defers to WebIDL, right?

index.bs
beaufortfrancois: I think `optional` is required. See https://chromium-review.googlesource.com/c/chromium/src/+/8250826/comment/43a901d5_d0bb54d8/

index.bs
beaufortfrancois: Without this, calling
```js
await document.modelContext.executeTool(tool);
// or
await document.modelContext.executeTool(tool, undefined);
```
would fail the following object check, and incorrectly reject with a TypeError instead of falling back to `{}`.

index.bs
beaufortfrancois: Done

index.bs
domfarolino: Ah, of course, we want the undefined case to obviously be treated like the empty object. In that case, see my comment on the Chromium CL. I'm hopeful we can just make the IDL `optional any inputObject = {}`.

domfarolino: Ah right, if we want to handle missing undefined and explicit undefined as both `{}` implicitly, I agree. In that case, I think we're better off expressing the IDL as `optional any inputObject = undefined`. If we don't, then I think we need to handle both explicitly here. See for example https://dom.spec.whatwg.org/#dom-abortcontroller-abort and other instances of `optional any`, which explicitly handle the "was given" case separate from just analyzing its type. This makes me think that Web IDL treats them distinctly. Anyways, let's resolve this over in https://crrev.com/c/8250826/comment/43a901d5_d0bb54d8/.

beaufortfrancois:
@domfarolino I've updated this PR based on the latest patchset uploaded in https://chromium-review.googlesource.com/c/chromium/src/+/8250826

domfarolino (approved):

PR #253, opened 2026-08-21
Add `debugging` member to ToolAnnotations

### LINKED ISSUE #207: Support developer tooling
One area of potential for WebMCP is to be a mechanism for developers of web apps, and developers of frameworks like React, Angular, etc, to provide debugging tools that can be used by developer tooling, such as Chrome DevTools AI Assistance and similar tools from other browsers. Additionally, any tool that can control a browser could call these tools.

### Debugging my web app's flickering UI
_Swap React here for any similar framework_

I’m building a React app and the UI of one of my components keeps flickering. I type into the input box and my mouse keeps losing focus.
I open my AI and ask “This UI keeps flickering and as I type into the input box my focus keeps getting hijacked. Debug this and find the problem”.
1. Because React has shipped their WebMCP debugging tools, DTAI can call componentForElement(‘.input-box’) to ask React to return information about the React component that generates that UI.
2. My agent decides that because the UI is flickering that likely means the component might be re-rendering too often and this could be a state issue. It calls `renderLog(InputComponent)` which is a WebMCP tool that React provides and exposes a log of the last 20 renders and their timestamps.
3. The agent confirms its suspicion as all the last 20 renders of the component are within the last few milliseconds, confirming that something in the component is triggering aggressive re-renders.
4. At this point, if the agent has source code access, it can look into the code to find a fix, but if it doesn't, it can still report a summary of its findings and the evidence to be passed onto a developer or another AI agent.

### Debugging my web app's state

I've got an app that tracks state in some state management tool (MobX, NgRx, etc). I'm noticing bugs where suddenly the state seems to change unexpectedly and become invalid.

1. I tell the agent the series of steps that are needed to reproduce (type in this field, click this button, etc).
2. As the agent does this, it can call the `logState()` WebMCP tool at regular, repeated intervals.
3. A tool is also exposed that can log the diff between states and the actions taken.
4. With this information, the AI can determine what message is erroneously being sent, causing the state to become incorrect. Or it can confirm that the messages are accurate, meaning the bug must exist in the processing of the messages.

## Previous art

Chrome DevTools for agents has shipped [support for third party developer tools](https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/third-party-developer-tools.md) using an API that is broadly similar to the WebMCP spec. If WebMCP is interested in supporting developer tooling, then long term this can migrate to support WebMCP rather than another distinct API.

## Requirements

The most obvious thing we will need is some way to distinguish between developer tools and user facing tools, so that an agent can filter its set of tools down - e.g. Gemini in Chrome does not want to call these tools, but DevTools AI might.

beaufortfrancois:
Following the [WebML CG resolution](https://www.w3.org/2026/08/20-webmachinelearning-minutes.html#73b8), this PR adds a `debugging` boolean member to the `ToolAnnotations` dictionary to provide a way for web apps and frameworks to register tools intended specifically for developer tooling and inspection (e.g., Chrome DevTools AI assistance, testing frameworks) so that general-purpose/end-user agents can filter them out.

Note that I went with `debugging` rather than `debuggingHint` because it is a definitive categorization flag identifying the tool's intended domain/audience (developer/debugging tools vs. user-facing tools) rather than a behavioral hint about how the tool executes.

Fixes #207

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/webmcp/pull/253.html" title="Last updated on Sep 11, 2026, 11:20 AM UTC (c768a6b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/253/97da8f5...beaufortfrancois:c768a6b.html" title="Last updated on Sep 11, 2026, 11:20 AM UTC (c768a6b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -211,6 +211,9 @@ An <dfn>annotations</dfn> is a [=struct=] with the following [=struct/items=]:
 
   : <dfn>untrusted content hint</dfn>
   :: a [=boolean=], initially false.
+
+  : <dfn>debugging</dfn>
+  :: a [=boolean=], initially false.
 </dl>
 
 <h3 id="pending-tool-executions">Pending tool executions</h3>
@@ -758,6 +761,9 @@ The <dfn method for=ModelContext>registerTool(<var>tool</var>, <var>options</var
       : [=annotations/untrusted content hint=]
       :: |tool|'s {{ModelContextTool/annotations}}'s {{ToolAnnotations/untrustedContentHint}}
 
+      : [=annotations/debugging=]
+      :: |tool|'s {{ModelContextTool/annotations}}'s {{ToolAnnotations/debugging}}
+
    : [=tool definition/exposed origins=]
    :: |exposed origins|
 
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
FYI I have a WIP Chromium CL at https://chromium-review.googlesource.com/c/chromium/src/+/8398587

markafoltz (approved):

markafoltz:
PR LGTM.

One question that comes to mind is whether a tool marked as `debugging` must not set `consequentialHint`.    Those two categories seem mutually exclusive, as debugging should not have irreversible, real-world side effects.

domfarolino (approved):

PR #258, opened 2026-08-25
[codex] Add ChatGPT Desktop to implementation status

nmccormack-oai:
## Summary

- Add ChatGPT Desktop to the WebMCP implementation-status list.
- Update the introduction to cover browser and client implementations.
- Check in the OpenAI Blossom SVG so the status page does not depend on a browser-only logo repository.

## Why

- ChatGPT Desktop supports WebMCP but is not represented on the implementation-status page.
- The existing logo source only hosts browser logos, so keeping the OpenAI asset in this repository avoids a cross-repository dependency and a separate upstream contribution.

## Validation

- Ran `git diff --check`.
- Parsed `assets/openai.svg` with `xmllint --noout`.
- Confirmed the SVG path comes from the official [OpenAI logo package](https://cdn.openai.com/brand/openai-logos.zip); only its intrinsic size and viewport were adjusted.
- Confirmed the fork branch and local branch point to the same commit.


victorhuangwq:
cc:// @beaufortfrancois and @domfarolino

I marked it as non-substantive as per https://www.w3.org/policies/process/#substantive-change since it does not touch any of the spec text.

implementation-status.md
```diff
@@ -2,14 +2,18 @@
...
+<a href="#brave"><img width=64 src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/brave/brave_128x128.png" alt="Brave logo"></a> <a href="#chatgpt-desktop"><img width=64 src="assets/openai.svg" alt="OpenAI logo"></a> <a href="#chrome"><img width=64 src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/chrome/chrome_128x128.png" alt="Chrome logo"></a> <a href="#edge"><img width=64 src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/edge/edge_128x128.png" alt="Edge logo"></a> <a href="#firefox"><img width=64 src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/firefox/firefox_128x128.png" alt="Firefox logo"></a> <a href="#safari"><img width=64 src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/safari/safari_128x128.png" alt="Safari logo"></a>
```
domfarolino: I'm a little torn on whether we want to link directly to an official asset hosted somewhere vs. check in the actual blossom asset, but I don't feel strongly, so I think this is fine.

domfarolino (approved):
It might be worth noting that on dark mode, the logo blends in a bit with the background. I suppose that's OK, but just flagging this in case you know of a better official logo that's resilient to this.

Otherwise, I'll plan on landing this tomorrow!

Image

anssiko (approved):

nmccormack-oai:
@domfarolino - great catch! Let me dig out another logo 😂

nmccormack-oai:
@domfarolino - should be good now!

domfarolino (approved):

Issue #182, opened 2026-05-22
How to expose failed declarative tool registration


domfarolino:
Tool registration can fail for multiple reasons: duplicate name, invalid name, permission policy is disabled etc. The imperative API handles this gracefully (usually by throwing `NotAllowedError` DOMExceptions, but how should we handle this for the declarative API? In a queued task, we should probably fire a `toolerror` event or some such, against either the offending `<form>` or the `ModelContext` object. I don't have strong opinions on which.

Or is there a better way to handle this?

anssiko:
(@domfarolino please re-apply Agenda+ when you feel it is appropriate time to have a group discussion.)

PR #259, opened 2026-08-25
Drop closed executeTool issue links from explainers

### LINKED ISSUE #51: Define the API for in-page Agents to use a site's declared tools

HaoChiBao:
PR #226 closed #51, #57, #117, #159, and #160. The explainers still listed those as open questions.

This drops the closed links and points the spec at declarative-api-explainer.md instead of PR #76.

#178 is still open; I left it out of the iframe "See:" list because the remaining ask is deployer guidance, not getTools()/executeTool(). #182 is unchanged.

***
<a href="https://pr-preview.s3.amazonaws.com/HaoChiBao/webmcp/pull/259.html" title="Last updated on Aug 25, 2026, 10:39 PM UTC (bdb74b7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webmcp/259/fca7462...HaoChiBao:bdb74b7.html" title="Last updated on Aug 25, 2026, 10:39 PM UTC (bdb74b7)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,12 +311,7 @@ The reason WebMCP is not limited to only declarative form tools is for the same 
 
 ### Permissions policy and iframes
 
-While much of this explainer assumes integration with built-in browser agents, WebMCP also supports **author-provided agents**, such as agents embedded directly on a page or running in an iframe, that can collaborate with parent frames and nested contexts. See:
- - [Issue #57](https://github.com/webmachinelearning/webmcp/issues/57)
- - [Issue #117](https://github.com/webmachinelearning/webmcp/issues/117)
- - [Issue #159](https://github.com/webmachinelearning/webmcp/issues/159)
- - [Issue #160](https://github.com/webmachinelearning/webmcp/issues/160)
- - [Issue #178](https://github.com/webmachinelearning/webmcp/issues/178)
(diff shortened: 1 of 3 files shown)


domfarolino (approved):
Thanks, much appreciated!

PR #279, opened 2026-09-01
Clarify executeTool input and schema values

### LINKED ISSUE #278: Clarify executeTool argument encoding and returned schema shape
While testing WebMCP tool registration and invocation in a browser implementation, I hit two API details that were easy to misread from the current examples/spec text.

Observed behavior:

```js
const tools = await document.modelContext.getTools();
const tool = tools.find((entry) => entry.name === "example_tool");

await document.modelContext.executeTool(tool, { message: "hello" });
// Fails with an input parsing / type error in the tested runtime.

await document.modelContext.executeTool(tool, JSON.stringify({ message: "hello" }));
// Succeeds in the tested runtime.

console.log(typeof tool.inputSchema);
// "string" in the tested runtime, containing serialized JSON schema.
```

The shape that worked was:

1. Await `document.modelContext.getTools()`.
2. Pass the `RegisteredTool` object returned by `getTools()` into `executeTool()`, not the tool name.
3. Pass JSON-encoded argument text into `executeTool()`.
4. Parse returned `inputSchema` metadata when it is exposed as a serialized JSON string.

It would help implementers if the explainer/spec/docs explicitly state whether `executeTool` arguments are intended to be a JavaScript object, a JSON string, or implementation-dependent for now, and whether returned tool schemas are expected to be structured objects or serialized strings.

If the current observed behavior is the intended contract, a small normative example for `executeTool` would prevent a class of false-negative debugging sessions where tools are registered correctly but every manual execution attempt appears broken.

mlmrx:
## Summary

- clarify that `executeTool()` takes a `RegisteredTool` returned by `getTools()`
- clarify that `RegisteredTool.inputSchema` is exposed as a parsed JavaScript object
- clarify that callers pass the structured JavaScript input value directly rather than pre-serializing it with `JSON.stringify()`
- add a focused discovery-and-execution example to the API introduction

This documents the contract already defined by the `RegisteredTool` dictionary and the
`executeTool()` serialization algorithm. An implementation that exposes `inputSchema` as a string or
requires pre-serialized JSON arguments does not match the current draft.

Closes #278.

## Validation

- W3C spec-generator: Bikeshed processing with `die-on=warning`
- `git diff --check`

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -635,8 +635,28 @@ is a [=model context=] [=struct=] created alongside the {{ModelContext}}.
   <dd>
     <p>Executes a tool on the document it was registered on. Returns a promise that resolves to the
     stringified result of the tool's execution.</p>
+
+    <p><var ignore>tool</var> is a {{RegisteredTool}} returned by {{ModelContext/getTools()}}. Its
+    {{RegisteredTool/inputSchema}} is exposed as a parsed JavaScript object. Callers pass the
+    structured JavaScript value described by that schema as <var ignore>inputObject</var>; the user
+    agent serializes and transfers it to the tool's execution context, so callers do not
+    {{JSON/stringify()}} it first.</p>
   </dd>
 </dl>
+
+<div class=example id=executing-a-discovered-tool>
+  The following discovers a tool and executes it with a JavaScript object:
+
+  ```js
+  const tools = await document.modelContext.getTools();
+  const tool = tools.find(({name}) => name === "example_tool");
+
+  if (tool) {
+    console.log(tool.inputSchema.type); // "object"
+    const result = await document.modelContext.executeTool(tool, {message: "hello"});
+  }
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -635,9 +635,29 @@ is a [=model context=] [=struct=] created alongside the {{ModelContext}}.
...
+    {{JSON/stringify()}} it first.</p>
   </dd>
 </dl>
 
+<div class=example id=executing-a-discovered-tool>
+  The following discovers a tool and executes it with a JavaScript object:
+
+  ```js
+  const tools = await document.modelContext.getTools();
+  const tool = tools.find(({name}) => name === "example_tool");
```
beaufortfrancois: Nit:

```suggestion
  const tool = tools.find(t => t.name === "example_tool");
```

Issue #268, opened 2026-08-28
[bug] Working on Brave, not working on MS Edge and Chrome


thiagodp:
Hello,

the following screenshot shows the "Application" tab and "WebMCP" for the same application, using, in order:

- Brave 1.94.117 [Chromium: 152.0.7977.64 (64 bits)]
- Chrome 152.0.7977.65 (64 bits)
- Edge 151.0.4129.107 (64 bits) 

<img width="1053" height="879" alt="Image" src="https://github.com/user-attachments/assets/3b2aa7b7-543b-4248-935e-e5597f9df8b5" />

Only on Brave the registered tools appear. So that's probably a bug. 

Using `"@mcp-b/global": "^5.0.1"`.

victorhuangwq:
Hey, just curious, have you enabled `about://flags/#enable-webmcp-testing` for both Edge and Chrome? 

It needs be enabled for your website to be able to register the tools, and thus for it to show up in the devtools. From what I know this isn't enabled by default yet for these two browsers.

thiagodp:
Haha, that was it. Thanks, @victorhuangwq.
This could be remembered on the project home page.

PR #280, opened 2026-09-01
Add flag for local development in Chrome

beaufortfrancois:
Following https://github.com/webmachinelearning/webmcp/issues/268#issuecomment-5465612565, this PR adds flag for local development in Chrome, which also works for Edge.

### DIFF
--- a/implementation-status.md
+++ b/implementation-status.md
@@ -18,6 +18,8 @@ WebMCP is supported in [ChatGPT Desktop](https://chatgpt.com/download/).
 
 An [Origin Trial](https://developer.chrome.com/blog/ai-webmcp-origin-trial) is live in Chrome 149.
 
+For [local development](https://developer.chrome.com/docs/ai/webmcp#local_webmcp), enable the `about:flags#enable-webmcp-testing` flag.
+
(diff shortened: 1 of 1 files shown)


implementation-status.md
```diff
@@ -18,6 +18,8 @@ WebMCP is supported in [ChatGPT Desktop](https://chatgpt.com/download/).
 
 An [Origin Trial](https://developer.chrome.com/blog/ai-webmcp-origin-trial) is live in Chrome 149.
 
+For [local development](https://developer.chrome.com/docs/ai/webmcp#local_webmcp), enable the `about:flags#enable-webmcp-testing` flag.
```
domfarolino, line 21: This just redirects to the more commonly known `chrome://flags[foo]`. Should we just make that explicit?

beaufortfrancois: I use `about:flags` because it works for all chromium-based browsers, and Edge actually refers to Chrome section.

domfarolino: Ah, thanks for pointing out that Edge refers to this section. OK that's fair enough I suppose.

domfarolino (approved):

PR #281, opened 2026-09-01
Include origin in observation tool map via observed tool collection struct

johannhof:
Discussed with @domfarolino offline - we want to add this to help track tool provenance for user agents outside of non-normative wording. Note that it's technically unused right now, we could follow up with standardizing serialization of tool origin + name. But that might be a more controversial change given [existing guidance in the spec](https://webmachinelearning.github.io/webmcp/#observation:~:text=Despite%20the%20name%20of%20this%20API%20(i.%2C%20WebMCP)%2C%20this%20specification%20does%20not%20prescribe%20the%20format%20in%20which%20tools%20are%20exposed%20to%20the%20browser%20agent.).


johannhof:
@domfarolino for review

index.bs
domfarolino: I suppose we should get rid of this, since it's explicit now?

domfarolino (approved):

index.bs
johannhof: Yeah maybe we should reframe this at least as a note for what the origin is meant for...

domfarolino (approved):

Issue #293, opened 2026-09-04
Two agent-browser behaviours a page cannot detect or recover from


webmyc:
Feedback from running a WebMCP integration in production against ChatGPT's in-app browser since 2026-08-26, on a live WordPress site with real users. Two findings where the page has no way to detect or recover from what the agent browser did.

### 1. The per-page tool budget fails silently and totally

Registering a large catalog (296 tools) disabled WebMCP for the entire document. No exception, no console warning, no partial registration. The page believes it succeeded. A curated set of roughly 30 tools works.

From the page's side those two outcomes are indistinguishable at registration time, so there is no way to feature-detect the ceiling or degrade toward it. A page that grows its tool set over time crosses the line at some point and simply stops being agent-accessible, with nothing in the page to indicate it.

Worth considering: a budget the page can read, or an ordering hint so a page over the limit loses its least important tools rather than all of them. The limit itself seems reasonable. Silent total disablement is the part that seems wrong regardless of where the limit sits.

### 2. `provideContext` and `registerTool` diverge, and the object is frozen

ChatGPT's browser exposes `document.modelContext` as a frozen object implementing only `registerTool`. Batch `provideContext` throws, and because the object is frozen, assigning a polyfill onto it silently no-ops in non-strict code.

The result is that an implementation written against the batch API registers zero tools while every other signal says discovery succeeded: the fetch resolved, the tool list parsed, no error surfaced. This cost us a day to find, and the fix was to prefer `registerTool` in a loop and fall back to `provideContext`.

If the spec has a view on which of the two is required rather than optional, saying so explicitly would let implementations pick correctly instead of discovering it against one browser.

### Offer

Happy to turn either into an eval in `webmcp-evals` if that shape is useful. The tool-count one is mechanical: register N tools, ask the agent to list them, walk N up until the answer becomes zero.

Context: the integration is [Respira for WordPress](https://github.com/respira-press/webmcp-for-wordpress) (GPL-2.0), running at https://chatgpt-respira.instawp.site. WordPress-specific integration traps from the same work are written up in [WordPress/ai#448](https://github.com/WordPress/ai/issues/448#issuecomment-5450915007).

domfarolino:
> Registering a large catalog (296 tools) disabled WebMCP for the entire document.

[...]

PR #294, opened 2026-09-04
Add Best Practices section and guidance on dynamic tool registration to README

beaufortfrancois:
Following https://github.com/webmachinelearning/webmcp/issues/293#issuecomment-5539780216, this PR adds a "Best Practices" section and guidance on dynamic tool registration to the README file.


domfarolino (approved):
So fast, thanks!! I'm wondering if we should include any of this guidance in the spec too. Not directly, but as a section that points out to this, in case developers are looking for it in the spec. Not sure that's necessary though, so happy to land this.

